PR #103, opened 2013-11-19
Fix problem with block blob creation from IO object.

allomov:
If you try to create block blob from `IO` object, you will have errors because IO doesn't have `bytesize` method and its md5 is counted another way.

``` ruby
content_file = File.open(image_file, 'r') do |io|
  azure_blob_service.create_block_blob(image_container_name, image_blob_name, io)  
end
```


allomov:
@guangyang ping

PR #133, opened 2014-03-18
do not pollute root namespace

tandibar:
I am using the Azure gem in a project and had some wired problems defining a `Location` class. After some investigation I found out that the class clashes with the Azure::BaseManagement::Location class. This is because of the line `include Azure::BaseManagement` in `lib/azure/base_management/base_management_service.rb`. This is really bad as it messes with the root namespace. I removed the `include` and fixed all references. All tests are passing. I did the same with the `Loggerx` and the include of the core utils.

### DIFF
--- a/lib/azure/base_management/base_management_service.rb
+++ b/lib/azure/base_management/base_management_service.rb
@@ -23,7 +23,7 @@ require 'azure/base_management/location'
 require 'azure/base_management/affinity_group'
 
 include Azure::Core::Utility
-Loggerx = Azure::Core::Logger
+Azure::Loggerx = Azure::Core::Logger
 
 module Azure
   module BaseManagement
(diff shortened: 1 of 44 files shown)


devigned:
@tandibar and @mukaibot I'm sorry this didn't receive more attention previously.

I agree with the premise of the pull request. We should not pollute the root namespace as it can lead to nastiness up the dependency chain.

With that said, if you'd like to rebase and freshen this up, I'd reevaluate merging the PR.

tandibar:
I merged current master with my changes. Hope you can accept this now. thx

devigned:
LGTM

tandibar:
Cool!

devigned:
Thank you for the contribution!

PR #141, opened 2014-04-29
Support for multiple configurations and a few other items

xplode:
What:  Added support for multiple configurations.
Why:  An application needs to manage multiple accounts in parallel.  The singleton configuration design prevents this.  I've made changes to Azure::Core::Configuration and classes that rely on this object to allow for both a single global configuration and configurations that can be passed into the azure api as needed.

What:  Added the ability to retrieve keys for a storage account.
Why:  An application needs to manage storage accounts without having the keys preloaded.  The keys must be retrieved via the Azure api.

What:  Included the media_link value in retrieved virtual machine images.
Why:  An application is creating virtual machines from account images.  The media_link info associated with these images contains the storage account name.  This is needed when creating the virtual machine from the image.

What:  Added ability to ignore case when retrieving virtual machines.
Why:  The Azure cloud ignores case on virtual machine and cloud service names.  You cannot create two machines named 'onevm' and 'oneVm'.  These are considered to be equivalent names.  The retrieval of a virtual machine by name should allow for the same equivalency.

### DIFF
--- a/lib/azure/base_management/base_management_service.rb
+++ b/lib/azure/base_management/base_management_service.rb
@@ -29,11 +29,12 @@ Loggerx = Azure::Core::Logger
 module Azure
   module BaseManagement
     class BaseManagementService
-      def initialize
+      def initialize(config=nil)
+        @config = config || Azure.config unless @config
         validate_configuration
-        cert_file = File.read(Azure.config.management_certificate)
+        cert_file = File.read(@config.management_certificate)
         begin
-          if Azure.config.management_certificate =~ /(pem)$/
+          if @config.management_certificate =~ /(pem)$/
             certificate_key = OpenSSL::X509::Certificate.new(cert_file)
             private_key = OpenSSL::PKey::RSA.new(cert_file)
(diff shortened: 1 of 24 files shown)


devigned:
I like the support for multiple configurations, but there are too many things in this pull request. If you break it up, we can evaluate and merge them.

devigned:
I would love to get this functionality in. If you want to update this, I'll get it pulled in.

coliver:
@devigned: Can you be more specific on how you would like this broken up? I, for one, really need this change and I'm sure others want it as well.

devigned:
When a pull request is added with a title that states it did one thing and a few others, that's a clear sign that it should be broken up into one pull request and a few other pull requests. 

To be more specific, the changes to configuration should be broken up from the changes for ignore_case and media_link.

djberg96:
Definitely would like to see these changes integrated as well, but can understand why David wants them broken up.

devigned:
Added in a large refactor: #248

PR #146, opened 2014-05-16
changed virtual machine management and image management to use vmimages ...

johnsmyth:
Changed virtual machine management and image management to use vmimages in addition to os images. Also added redirect code for http 307 response codes. I made these changes because we are no longer able to use the azure gem (or vagrant azure) to create new VMs from custom images that we create. I believe there may have been a change this month where user vmimages that are captured are treated differently than standard OS images. This code seems to work, but it took a lot of trial and error as I couldn't find very good documentation. I'm happy to do some further refining of this code if I were pointed to the documentation detailing the differences between /services/images and /services/vmimages, as well as the proper deployment options.

### DIFF
--- a/lib/azure/base_management/management_http_request.rb
+++ b/lib/azure/base_management/management_http_request.rb
@@ -36,7 +36,7 @@ module Azure
         @warn = false
         content_length = body ? body.bytesize.to_s : "0"
         @headers = {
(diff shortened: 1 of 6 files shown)


devigned:
Should be fixed in 0.7.0.pre

PR #150, opened 2014-06-09
Expose private IP for VMs

mukaibot:
Allow the VM's private IP address to be accessed as a property.


FestivalBobcats:
:+1: extremely important for clustered databases... guess I'll be using the fork.

devigned:
Clearly, this PR is way out of date, but the functionality should be in this repo. I'll make sure this functionality is added.

Issue #153, opened 2014-06-24
Cannot set up management_certificate to a .pfx file


nicolasdao:
Setting the Azure configuration with a .pem file works fine, but once I've switched to its pfx counterpart as follow:

```ruby
Azure.configure do |config| 
      config.storage_account_name = "my_storage_name"
      config.storage_access_key = "my_storage_key"
      config.management_certificate = "my_cert.pfx"
      config.subscription_id = "my_subscription_id"
      config.management_endpoint = "https://management.core.windows.net"
end
```
Then I receive the following error message:
```ruby
Management certificate not valid. Error: PKCS12_parse: mac verify failure
```

devigned:
Should be fixed in 0.7.0.pre -- check out pfxer in the bin dir.

PR #154, opened 2014-06-24
#153 - Fix bug related to passing .pfx file to management_certificate rather than .pem file

nicolasdao:
Fix issue #153.

#### Fix Details
Though no issues were encountered when using a .pem file, a parsing issue was occurring with .pfx file. The reason was that the same file reading mechanism was used on both file types, though their encoding is different(plain text for .pem and binary for .pfx). The solution simply consisted in opening the file differently based on on their extension

### DIFF
--- a/lib/azure/base_management/base_management_service.rb
+++ b/lib/azure/base_management/base_management_service.rb
@@ -31,14 +31,18 @@ module Azure
     class BaseManagementService
       def initialize
         validate_configuration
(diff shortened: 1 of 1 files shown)


devigned:
@nicolasdao if you add a test to verify this functionality, I'll get it pulled in.

devigned:
Should be fixed in 0.7.0.pre

Issue #155, opened 2014-06-24
timeout on Azure::Blob::BlobService - create_block_blob doesn't allow the client to not timeout after more than 2 min.


nicolasdao:
I receive a timeout error when I try to upload a file that requires more than 2 minutes(default timeout on Net::HTTP) before its upload is completed. I was hoping to be able to set up the client's timeout by doing the following:

```ruby
blob_options = {
	:content_type => "application/octet-stream",
	:timeout => 15*60 # 15 minutes
}
blob_srv = Azure::Blob::BlobService.new
new_blob = blob_srv.create_block_blob("my_container", "some_name", blob_content, blob_options)
```

But it looks like it only affects the Azure timeout, but not the Net::HTTP client since I keep having the timeout error after 2 minutes. It would be nice if that :timeout setting would affect both.

devigned:
Should be fixed in 0.7.0.pre

PR #156, opened 2014-06-24
#155 -  Fix issue related to timeout only setting the azure timeout but not the Net::HTTP client

nicolasdao:
Fix issue #155.

#### Fix Details
When specifying the timeout for method create_block_blob as follow:
```ruby
blob_options = {
    :content_type => "application/octet-stream",
    :timeout => 15*60 # 15 minutes
}
```
A "Keep-Alive: timeout=900" is added to the HTTP header. This is later extracted inside the http_request.rb to determine whether or not the Net::HTTP client should use its default timeout or replace it with the specified value(in our example, 900)


devigned:
Should be fixed in 0.7.0.pre

Issue #157, opened 2014-06-24
Cannot delete the staging deployment of a cloud service


nicolasdao:
Missing optional parameter 'slot' on method 'delete_cloud_service_deployment' in file cloud_service_management_service.rb. The current implementation has hardcoded the deployment to 'production'(cf. line 124). 

The proposed solution is to add an optional parameter called slot that is defaulted to 'production' as follow:
```ruby
def delete_cloud_service_deployment(cloud_service_name, slot='production')
  slot = "production" unless slot
  request_path = "/services/hostedservices/#{cloud_service_name}/deploymentslots/#{slot}"
  request = ManagementHttpRequest.new(:delete, request_path)
  Loggerx.info "Deleting deployment of cloud service \"#{cloud_service_name}\" ..."
  request.call
end
```

devigned:
Should be fixed in 0.7.0.pre

PR #158, opened 2014-06-24
#157 -  Fix issue related to the inability to delete a staging deployment in a cloud service

nicolasdao:
Fix issue #157.

#### Fix Details
Implement the solution proposed in #157


devigned:
Should be fixed in 0.7.0.pre

PR #159, opened 2014-06-26
Add 3 new web hooks inside "cloud_service_management_service.rb":

nicolasdao:
- create_deployment (based on microsoft doc at http://msdn.microsoft.com/en-us/library/azure/ee460813.aspx)
- upgrade_deployment (based on microsoft doc at http://msdn.microsoft.com/en-us/library/azure/ee460793.aspx)
- get_deployment (based on microsoft doc at http://msdn.microsoft.com/en-us/library/azure/ee460804.aspx)

For each of those methods, a unit test has been added inside the 'cloud_service_management_service_test.rb'


devigned:
this is out of date. I'm going to close it.

PR #166, opened 2014-08-22
My images

nicholasjackson:
Added capability to retrieve image information for images which have been user created from existing virtual machines.


devigned:
Should be fixed in 0.7.0.pre

PR #176, opened 2014-10-09
Allow IO as body

nurse:
For example BlobService#create_block_blob says its content is
IO or String but actually doesn't.


devigned:
I'd really like to see this get pulled in. Would you be able to update this pr?

nurse:
@devigned updated!

devigned:
Added in a large refactor: #248

Issue #171, opened 2014-10-02
Missing support for new VM sizes


dJason:
The A8, A9, and newly introduced D* series are missing from the allowed VM sizes.

sepiroth887:
Would it be possible to cherrypick this in and add the new G size machines as well?? Also a release would be nice in order for the vagrant-azure repo to leverage this change?

devigned:
#217 should fix this up. I'll leave it open until gem release.

devigned:
Should be fixed in 0.7.0.pre

PR #179, opened 2014-11-03
Fixed issue: Missing support for new VM sizes #171

ranjan:


devigned:
The role sizes stuff is great, the private IP stuff is awesome, the injection of custom data is fantastic, but please don't submit a mass PR for all of them without even a description. We'd would love to get these features in, but please submit each feature independently.

I'm going to close this PR, but please resubmit independently.

PR #191, opened 2015-01-21
Added missing validate_configuration methods + Fixed tests

mstuefer:


lib/azure/blob/blob_service.rb
```diff
@@ -21,6 +21,7 @@ module Blob
     class BlobService < Service::StorageService
 
       def initialize
+        validate_configuration
         super()
```
devigned: Would you drop this call to super?

lib/azure/queue/queue_service.rb
```diff
@@ -20,6 +20,7 @@ module Queue
     class QueueService < Service::StorageService
 
       def initialize
+        validate_configuration
         super()
```
devigned: Would you drop this call to super?

lib/azure/service/storage_service.rb
```diff
@@ -21,8 +21,9 @@ class StorageService < Azure::Core::SignedService
       # Create a new instance of the StorageService
       #
       # signer        - Azure::Core::Auth::Signer. An implementation of Signer used for signing requests. (optional, Default=Azure::Core::Auth::SharedKey.new)
-      # account_name  - String. The account name (optional, Default=Azure.config.storage_account_name)  
+      # account_name  - String. The account name (optional, Default=Azure.config.storage_account_name)
       def initialize(signer=Core::Auth::SharedKey.new, account_name=Azure.config.storage_account_name)
+        validate_configuration
         super(signer, account_name)
```
devigned: Would you drop this call to super?

lib/azure/version.rb
```diff
@@ -17,7 +17,7 @@ module Azure
   class Version
     MAJOR = 0 unless defined? MAJOR
     MINOR = 6 unless defined? MINOR
-    UPDATE = 4 unless defined? UPDATE
+    UPDATE = 5 unless defined? UPDATE
```
devigned: This is updated in master.

devigned:
Please update this PR to be current with master.

PR #197, opened 2015-02-08
Fix 'add_role' unit test launched without 'turn' gem.

valtri:
One unit test fails when 'turn' gem is not available:

Azure::VirtualMachineManagement::VirtualMachineManagementService::#add_role#test_0003_throws error when certificate path is not invalid. [/usr/share/gems/gems/azure-0.6.4/test/unit/virtual_machine_management/virtual_machine_management_service_test.rb:426]:
Expected /No such file or directory -/i to match "\e[31m\e[1mNo such file or directory @ rb_sysopen - f:/invalid_path/private_key\e[0m\e[0m".

Fixed by adding '*' as used in all other tests.


devigned:
fixed via #214

PR #198, opened 2015-02-18
Allow access for the new VM sizes

zeitiger:
Allow access to the new VM sizes D*, DS* and G*
https://msdn.microsoft.com/en-us/library/azure/dn197896.aspx


marcy-terui:
:+1:

stuartpreston:
Great to see this merged... as an additional suggestion how about instead of restricting the VM role sizes, send a warning if it isn't one of the 'known' machine sizes, but still attempt to create the machine with the specified size anyway.  
This would avoid the SDK lagging behind new releases of machine sizes.

devigned:
:+1: pr me asap :)

PR #207, opened 2015-03-11
BlobService: add ms-blob-content-disposition support, bump base `x-ms-version` header to `2013-08-15`

rahim:
The version bump in `Azure::Core::Http:HttpRequest` is potentially breaking, but required to support the new field.

If it is breaking the bumped headers could be extracted and isolated to `BlobService` or only the affected methods.

### DIFF
--- a/lib/azure/blob/blob_service.rb
+++ b/lib/azure/blob/blob_service.rb
@@ -29,25 +29,25 @@ module Azure
       #
       # ==== Attributes
       #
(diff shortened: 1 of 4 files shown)


devigned:
I want to get this merged. Please add an integration test to the blob properties to verify the functionality.

devigned:
Should be fixed in 0.7.0.pre

PR #209, opened 2015-03-19
Blob service shared access signatures

rahim:
I still intend to add convenience methods for consistency on BlobService itself, but this is useable as is.


rahim:
Ignore my comment above - I'm not convinced adding convenience methods on BlobService would be helpful, but happy to add if anyone thinks they are.

devigned:
Please add an integration test showing it's usage via BlobService.

devigned:
@rahim thank you for the great work

rahim:
Pleasure! Apologies for not replying regarding the integration test, I'd been meaning to address that following your comment but I'd not yet found time.

devigned:
No worries. You waited much longer for your reply on the pr :smile:

Issue #210, opened 2015-03-23
Security issue: SSL certificate is not verified


bukk530:
I was wondering if the communication between my servers and azure was secured by SSL and I noticed something strange: https://github.com/Azure/azure-sdk-for-ruby/blob/master/lib/azure/core/http/http_request.rb#L146
It means the connction is encrypted but is susceptible to MITM attacks, as any tampered certificate would be accpted.
Am I just paranoid or I'm missing something?

devigned:
@bukk530 I don't think you're paranoid. Could you submit a PR?

bukk530:
I came up with this, and It's working fine https://github.com/Azure/azure-sdk-for-ruby/pull/211

devigned:
fixed in #219 and #220

PR #211, opened 2015-03-23
SSL certificate check

bukk530:
Now checking SSL certificate as described in the issue #210


devigned:
Are you running into on Windows?
Can you smash up the changes into 1 commit?

Also, I'm not sure how I feel about checking in the ca_bundle rather than using the system ca directory. I definitely think it's better than the alternative of using no peer verification. Thoughts?

devigned:
I'm thinking this path might be a little less prone to future issues: https://gist.github.com/fnichol/867550.

bukk530:
The problem about reading the ca from a system directory is that each system stores it in a different directory. However I've hardcoded the ca in the repository as I've seen doing this many times.
The strategy described here https://gist.github.com/fnichol/867550 would mean we have to create a generator that downloads and installs the ca each time one install this gem, am I right?
One last thing, instead of packing any trusted cert in the repository we could just use Microsoft ones, as done by Facebook in their PHP SDK: https://github.com/facebook/facebook-php-sdk-v4/blob/4.0-dev/src/Facebook/HttpClients/certs/DigiCertHighAssuranceEVRootCA.pem

devigned:
@bukk530 if you would narrow it down to just the MS related certs. I'll get this pulled in. Great work and thank you for the contribution.

devigned:
Closing based on #219 and #220. Thank you for the PR, but I don't think we should be in the business of distributing cert bundles.

PR #217, opened 2015-04-10
Allow all vm_sizes and warn if an unknown size

### LINKED ISSUE #216: Suggestion: Avoid SDK lagging behind new VM role sizes when announced
Instead of restricting the VM role sizes (ref #198), I suggest to send a warning if the VM role is isn't one of the 'known' machine sizes, but still attempt to create the machine with the specified size anyway.

This would avoid the SDK lagging behind new releases of VM role sizes.

Putting together a PR for this shortly...

stuartpreston:
Add A10 and A11 to valid role sizes...
But now only warn if a vm is not in the list of valid role sizes and allow Azure to decide whether it is valid or not at runtime.
Fixes #216

### DIFF
--- a/lib/azure/virtual_machine_management/virtual_machine_management_service.rb
+++ b/lib/azure/virtual_machine_management/virtual_machine_management_service.rb
@@ -522,15 +522,15 @@ module Azure
       end
 
       def validate_role_size(vm_size)
-        valid_role_sizes = %w(
+        suggested_role_sizes = %w(
           Basic_A0 Basic_A1 Basic_A2 Basic_A3 Basic_A4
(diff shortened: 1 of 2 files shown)


devigned:
:+1:  After CI passes, this is getting pulled in.

devigned:
@stuartpreston looks like a couple of the tests are sad. Can you assuage their sadness through some masterful codes?

stuartpreston:
Yes, working on it...  Having to reset my environment here to get MiniTest working.  Anything above 5.2.1 complains bitterly.

devigned:
yeah... I will be addressing that. Sorry for the pain.

Issue #203, opened 2015-02-23
Rescuing Exception class


arax:
Rescuing `Exception` class in Ruby is a big no-no. Please, have a look at [this post](http://blog.nicksieger.com/articles/2006/09/06/rubys-exception-hierarchy/) for more information. It is a critical issue.

See `lib/azure/base_management/base_management_service.rb` (among others)

``` ruby
begin
  # ...
rescue Exception => e
  raise "Management certificate not valid. Error: #{e.message}"
end
```

devigned:
Agreed. :+1:

If you'd like to submit a PR, we'll get it pulled in.

sarangan12:
@arax Please reopen this issue when you have a PR. Since this is an old issue, I am closing it

PR #231, opened 2015-05-08
Don't rescue exception, and use File.extname where appropriate

djberg96:
This replaces "rescue Exception" with "rescue OpenSSL::OpenSSLError" in base_management_service.rb. I also made some minor tweaks so that it uses File.extname instead of a regex, as well as ignoring case, to check for the extension.

https://github.com/Azure/azure-sdk-for-ruby/issues/203

### DIFF
--- a/lib/azure/base_management/base_management_service.rb
+++ b/lib/azure/base_management/base_management_service.rb
@@ -24,14 +24,14 @@ module Azure
         validate_configuration
         cert_file = Azure.config.management_certificate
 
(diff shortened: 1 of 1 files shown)


devigned:
Great improvement. We'll definitely pull this in.

djberg96:
Thanks! I see you have several PR's out there. Any idea when these will be merged?

devigned:
I think today would be good

PR #234, opened 2015-05-11
Missing ConfigurationSet for OS for custom images in request

jdimmerman:
Custom images are assigned an image_type of 'VM as compared to stock images, which have a type of 'OS'.

Without adding  || image.image_type == 'VM' at line 92 of serialization.rb, the serialized request does not contain a ConfigurationSet for the OS (only one for the network).


devigned:
Great catch!

Issue #225, opened 2015-04-30
Call to #update_endpoints and #delete_endpoint causes VM to loose StaticVirtualNetworkIPAddress


sangy14:
If a VM in a virtual network contains a static internal ip address, call to #update_endpoints and #delete_endpoint causes the VM to loose the static internal ip address.

Steps to reproduce:-
1>Create a VM in virtual network
2>Set a Static Internal IP Address for the VM using steps available on https://msdn.microsoft.com/en-us/library/azure/dn630228.aspx
3>Use the following snippet to delete endpoint
virtual_machine_service = Azure::VirtualMachineManagementService.new
virtual_machine_service.delete_endpoint('test-vnet', 'test-azure-vnet', 'Remote Desktop')

You can use either of the following steps to verify the issue
----------------------------------------------------------------------------------------------------------------------------
1>Use REST Client to call GET ROLE API. [You can use this link to for more info https://msdn.microsoft.com/en-us/library/azure/jj157193.aspx]
2>XML response doesn't contain \<StaticVirtualNetworkIPAddress\>ip-address\</StaticVirtualNetworkIPAddress\>

----------------------------------------------------------------------OR--------------------------------------------------
1>Shutdown VM
2>Start VM again. You will see changed internal ip address.

devigned:
Thank you for the steps for repro. We'll give it a look.

PR #236, opened 2015-05-14
#225 Call to #update_endpoints and #delete_endpoint causes VM to loos…

sangy14:
…e StaticVirtualNetworkIPAddress


devigned:
LGTM

Issue #208, opened 2015-03-18
Why raise a message but not a exception in VirtualMachineManagementService.create_virtual_machine?


AbelHu:
I want to get the error code of the error in VirtualMachineManagementService.create_virtual_machine but this method only raise a message. This is very bad. Could I change it to raise the exception?

Azure::VirtualMachineManagement::VirtualMachineManagementService.create_virtual_machine(...)
      rescue Exception => e
        e.message
      end

devigned:
Agreed. This should provide more feedback than just a string. Please submit a PR.

djberg96:
In this particular case it looks like it should just be removed, and the exception allowed to bubble up.

djberg96:
Actually, I guess to be consistent with the rest of the code base, it's better to use Azure::Logger.error_with_exit. I'll submit a PR.

djberg96:
Welp, that broke the test suite, but I think these are supposed to be expected, so it may be an issue with the tests themselves.

devigned:
fixed in: #246

PR #238, opened 2015-05-14
Re-raise exception in VirtualMachineManagementService#create_virtual_machine

djberg96:
On exception actually raise an error via Azure::Loggerx instead of swallowing the error and only returning a message. Plus, some formatting.

Note that this will cause some test failures. I think it is the tests that need updating.

Submitted in response to: https://github.com/Azure/azure-sdk-for-ruby/issues/208


devigned:
I like where this is going

djberg96:
Thanks. If you look at the test failures, the behavior would seem to match the test description, it just doesn't happen to pass the test. That's why I suspect the tests need updating.

devigned:
closed in favor of #246.

PR #244, opened 2015-05-25
VM create > change affinity_group_name selection order

ebuildy:
Better to select first the method parameters rather than the value of the virtual network (which is empty in my case, and make impossible the use of "create_virtual_machine" method).


devigned:
Would you be so kind and update this pull request.

PR #246, opened 2015-06-03
raise exception in case of vm creation failure

sangy14:
Raise exception if vm creation fails.


djberg96:
This obviates the need for my own PR since this one actually updated the tests as well. https://github.com/Azure/azure-sdk-for-ruby/pull/246.

Once you merge this one, close mine please.

PR #249, opened 2015-06-16
Adding simplecov coverage

khoale-azure:


devigned:
Would you be so kind and update this pull request. I'll get it pulled in when it is updated.

PR #255, opened 2015-07-22
Make the unit test pass on machines with F: for CD/DVD drive

slepox:
This test will fail on a machine with F: for CD/DVD drive.
The reason is the test assumed that when f:/invalid_path/ doesn't exist, the error message is 'No such file or directory ...'. However on a machine with F: as a CD/DVD drive, it doesn't allow to access a path if no CD/DVD is inserted, so the error message will be "... Permission denied...".

Just change to c:/invalid_path, c: will mostly be a system drive and not in such case.

### DIFF
--- a/test/unit/virtual_machine_management/virtual_machine_management_service_test.rb
+++ b/test/unit/virtual_machine_management/virtual_machine_management_service_test.rb
@@ -318,8 +318,8 @@ describe Azure::VirtualMachineManagementService do
     it 'throws error when certificate path is not invalid.' do
       options = {
           winrm_transport: %w(https http),
-          private_key_file: 'f:/invalid_path/private_key',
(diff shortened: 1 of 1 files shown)


azurecla:
Hi __@slepox__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (jiyang). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
        If you are a vendor, or work for Microsoft Open Technologies, DO please sign the electronic contribution license agreement. It will take 2 minutes and there's no faxing! https://cla.azure.com.
    </p>

TTYL, AZPRBOT;

PR #260, opened 2015-08-13
allow anything greater than 1 for mimetype so that vagrant-share works

devigned:


RocketPop:
Please merge and release this as soon as possible - if azure gem wouldn't force requirement of mime-type 2.0 I can stop hand-patching gems I use :)

devigned:
will do @RocketPop

Issue #266, opened 2015-09-02
Parse information from publishsettings file


andrewelizondo:
The majority of users will already have a publishsettings file, which contains all of the needed information. Instead of requiring additional steps/information why couldn't the publishsettings file be parsed for the information needed? Nokogiri is already a dependency of the service gem and could be used to parse the file.

andrewelizondo:
This was resolved in #271

PR #271, opened 2015-09-08
Allow for a publishsettings file to be provided for the certificate, solves #266

andrewelizondo:
This PR implements the ability for a user to provide a .publishsettings file to `management_certificate` and it will be parsed for the appropriate certificate, removing the need for an additional step to convert the .publishsettings file to a .pfx

This could potentially be extended later to parse for the subscription ID as well, but for now that hasn't been implemented here.


azurecla:
Hi __@andrewelizondo__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

devigned:
@andrewelizondo, nice! Could you add a unit test to ensure no future regressions?

andrewelizondo:
@devigned I got a fixture in place and unit tests added, I don't see a situation where someone would try to load this as a string so I didn't add that test, but I can be convinced.

devigned:
LGTM. Thank you, @andrewelizondo!

PR #273, opened 2015-09-10
Resolves #272 - BlobService::call fails when body is an IO 

### LINKED ISSUE #272: BlobService::call fails when body is an IO
Both the body.encoding and body.force_encoding in BlobService::call fail when body is an IO.

```ruby
      def call(method, uri, body=nil, headers=nil)
        # Force the request.body to the content encoding of specified in the header
        # (content encoding probably shouldn't be used this way)
        if headers && !body.nil?
          if headers['Content-Encoding'].nil?
            headers['Content-Encoding'] = body.encoding.to_s
          else
            body.force_encoding(headers['Content-Encoding'])
          end
        end
```

Am I missing something?  Want me to fix?

zachproffitt:
BlobService attempts to synchronize encoding between the Content-Encoding header and the body encoding.  There were two issues:

1) Body can be either a String of IO, but the code only worked with respect to Strings.
2) Azure expects Content-Encoding to be lowercase, otherwise authentication fails.

### DIFF
--- a/service_management/azure/lib/azure/blob/blob_service.rb
+++ b/service_management/azure/lib/azure/blob/blob_service.rb
@@ -1367,14 +1367,19 @@ module Azure
       end
 
       def call(method, uri, body=nil, headers=nil)
-        # Force the request.body to the content encoding of specified in the header
-        # (content encoding probably shouldn't be used this way)
+        # Synchronize body and header encoding; header['Content-Encoding'] takes precedence. 
         if headers && !body.nil?
           if headers['Content-Encoding'].nil?
-            headers['Content-Encoding'] = body.encoding.to_s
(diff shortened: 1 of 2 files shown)


azurecla:
Hi __@zachproffitt__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

devigned:
@zachproffitt I love it! Could you also add a unit test to ensure no regressions?

zachproffitt:
@devigned, I added tests to cover these scenarios.  Let me know if you have any other feedback.  Thanks!

devigned:
LGTM

devigned:
Thank you for the pr, @zachproffitt

PR #274, opened 2015-09-11
missing id attribute in network_interface_reference

johnsmyth:


azurecla:
Hi __@johnsmyth__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

devigned:
Hmm... I believe there should be an id attribute there based on the SubResource inheritance (see: https://github.com/Azure/autorest/blob/master/ClientRuntimes/Ruby/ms-rest-azure/lib/ms_rest_azure/sub_resource.rb#L11).

johnsmyth:
my bad

devigned:
No worries. Please don't hesitate with feedback or improvements.

johnsmyth:
Thanks.  What would really be helpful would be more documentation and/or sample code. If you could point toward docs or code, that would be awesome.  I'd really like to work on a Chef Provisioner driver for ARM, but I've been struggling trying to figure out how to use these new modules.

devigned:
Samples and docs are lacking and that is something we are working on... I don't want to ask too much, but do you have a list of goal scenarios you'd like to see in the samples. For example, "I need to set up a set of machines behind a load balancer" or some other end goals which we could then build out into a sample.

We are planning on building a series of these goal oriented samples, and provide an equivalent implementation across all of our SDKs. This SDK obviously would be included in those sample implementations.

If you have any ideas, we're all ears. We'll get them prioritized and start executing.

@stuartpreston, you have any suggestions?

stuartpreston:
Hi John, just saw your comment - were you aware of http://github.com/pendrica/chef-provisioning-azurerm ?

devigned:
Wow @stuartpreston, you have a great interweb connection. Mention you and 7 seconds later, you respond.

stuartpreston:
@devigned Incredible eh! I just happened to be catching up on my GitHub notifications and saw the post from @johnsmyth 

John, we would love any contributions you were able to make to the chef-provisioning-azurerm project. Wiring up more of the resources is next on my list but wanted to get basic deployment templates functionality out first.

johnsmyth:
@devigned,  I think some basic docs on expected behaviors, how to check for success/failure and handle errors and exceptions is important. 
Also, is there a way to create multiple dependent objects in a single create_or_update call (a vm, it's interfaces, extensions, etc)?  


@stuartpreston, I did see your gem and noticed the lack of a driver for supporting macine and machine_inage.  I was thinking about working on implementing that but If you have other priorities in mind I'm flexible.

stuartpreston:
@johnsmyth Yes, the route to machine and machine_image really needs to be via support for the discrete resources used. Unlike with ASM where we had a structured model with a cloud service being the container for the VMs, there's much more flexibility now. Which means we have to make some choices for what a 'machine' would look like. Does it have a public IP? If not how will it bootstrap? Do we need to put a load balancer in front of it in case we want to add more machines in future? Etc. Which is why I elected to start on the deployment template (which is the way to create multiple dependent objects, but obviously then requires JSON knowledge). Would be grateful for your thoughts and contributions.

PR #289, opened 2015-10-06
service is returning westus rather than West US

devigned:


djberg96:
To be honest, I was never entirely clear on how Azure interpreted locations. Seems like "West US" and "westus" (and other analogues) are basically equivalent from an API perspective, at least on the V2 side.

Issue #293, opened 2015-10-08
:udp_endpoints are not being created 


rinat-io:
Using SDK 0.7.1

```
*** LOCAL GEMS ***

azure (0.7.1)
```

Need to create VM with identical set of endpoints for both TCP & UDP protocols.
Added following 2 options to VM object definition:

```
      :tcp_endpoints => '3389, 8080, 14000, 14001, 14010, 14100, 14101, 14200, 15000, 15001, 15010, 15100, 15101, 15200, 16000, 16001, 16010, 16100, 16101, 16200, 22000, 22001, 24000, 24001, 25000, 25001, 26000, 26001',
      :udp_endpoints => '3389, 8080, 14000, 14001, 14010, 14100, 14101, 14200, 15000, 15001, 15010, 15100, 15101, 15200, 16000, 16001, 16010, 16100, 16101, 16200, 22000, 22001, 24000, 24001, 25000, 25001, 26000, 26001',
```

But for some reason only TCP endpoints were created.
Could some one confirm or correct why UDP endpoints were not created?

Thank you

veronicagg:
@rinat-io Thanks for catching this. It looks like serialization for udp_endpoints was missing in the sdk, I just submitted a pull request to add it. Will get it reviewed and resolved soon.

veronicagg:
@rinat-io This issue should be resolved now.

PR #336, opened 2016-03-23
Adding serialization for udp endpoints and corresponding tests.

veronicagg:
Addressing issue #293.


service_management/azure/lib/azure/virtual_machine_management/serialization.rb
```diff
@@ -99,6 +99,10 @@ def self.role_to_xml(params, image, options)
                     options[:tcp_endpoints],
                     options[:existing_ports]
                   ) if options[:tcp_endpoints]
+                  udp_endpoints_to_xml(
+                      xml,
+                      options[:udp_endpoints]
```
vishrutshah: should we also pass on options[:existing_ports] to udp_endpoints_to_xml method like we do in tcp_endpoints_to_xml?

vishrutshah:
LGTM!

Issue #291, opened 2015-10-06
Need to add recordings for tests


devigned:
The tests take too long and are really annoying to setup for contributors. The tests should have recording files setup so that they are relatively quick and easily approachable.

vishrutshah:
Merged all existing ARM & ASM VCR recordings to master.

PR #350, opened 2016-04-11
Adding recordings for ASM database integration tests

vishrutshah:
This changes is part of the feature #291
This also fixes following two issues in the ASM database SDK
1. Resetting database password was hitting the malformed URL because 'reset_password' method of sql_database_management_service.rb file at line 99 was concatenating string instead of calling method
2. When invalid location was given to create database, SDK was throwing 'No method found' in serialization.rb at line 60 while de-serializing response_xml.css('ServerName').first['FullyQualifiedDomainName'] because response_xml.css('ServerName') was not an array


azurecla:

TTYL, AZPRBOT;

service_management/azure/test/integration/test_helper.rb
```diff
@@ -35,7 +35,7 @@
   config.default_cassette_options = {:record => :once}
   config.cassette_library_dir = "vcr/vcr_cassettes"
   config.hook_into :faraday
-  config.allow_http_connections_when_no_cassette = false
+  config.allow_http_connections_when_no_cassette = true
```
veronicagg: do we need this, or should we just delete this line?

veronicagg:
@vishrutshah just made one more comment, rest looks good to me. Thanks!

service_management/azure/test/integration/test_helper.rb
vishrutshah: I believe that we need this line to make sure recorded tests never makes requests on network. We need this to be false. Thanks for catching this Fixing it on merge.

vishrutshah:
@veronicagg I've fixed the lazy initialization pattern and disables VCR to make network requests on no cassette. 
Thanks for the review !

Issue #360, opened 2016-04-13
Can not upload file with spaces


malibeg:
With new version 0.7.3. I can not upload files with spaces in filename. On azure file is created with content-length of zero and filename contains html encoding with %20

devigned:
Thank you for the issue report. We'll take a look at this asap.

@veronicagg or @vishrutshah, please take a look.

vishrutshah:
@malibeg & @devigned 

We have reproduced the problem and it seems like regressed from PR #303 We are looking into it why we are encoding the path twice.

vishrutshah:
@malibeg Thanks for reporting this. It'd be great if you'd like to review the PR for the fix, we'd appreciate it.

vishrutshah:
Hi @malibeg, we have fixed the problem and released a new version of azure gem v0.7.4. 

Please use [Azure Gem v0.7.4](https://rubygems.org/gems/azure/versions/0.7.4) and let us know your feedback, 

Thanks,

PR #361, opened 2016-04-13
Resolve double encoding problem

vishrutshah:
This is regarding the Issue #360 where users were uploading file with spaces in name then uploaded file has encodided characters e.g. %20 for space. This is due to the double encoding problem (once in [blob_service.rb:1442](https://github.com/Azure/azure-sdk-for-ruby/blob/master/service_management/azure/lib/azure/blob/blob_service.rb#L1442) and second time in [service.rb:40](https://github.com/Azure/azure-sdk-for-ruby/blob/master/service_management/azure/lib/azure/core/service.rb#L40)). 

This change removed encoding from [blob_service.rb:1442](https://github.com/Azure/azure-sdk-for-ruby/blob/master/service_management/azure/lib/azure/blob/blob_service.rb#L1442) and updated other CGI.escapse method to URI.encode method.


azurecla:

TTYL, AZPRBOT;

service_management/azure/lib/azure/service_bus/auth/shared_access_signer.rb
```diff
@@ -23,9 +23,9 @@ def name
         end
 
         def token(uri)
-          url_encoded_resource = CGI.escape(uri.to_s.downcase).gsub('+', '%20').downcase
+          url_encoded_resource = URI.escape(uri.to_s.downcase).downcase
```
veronicagg: I was reading about [the difference between uri escape and cgi escape](http://stackoverflow.com/questions/2824126/whats-the-difference-between-uri-escape-and-cgi-escape), have you looked into this? Maybe we should not be using URI.escape?

vishrutshah:
@veronicagg Thanks for referencing "[What's the difference between URI.escape and CGI.escape?](http://stackoverflow.com/questions/2824126/whats-the-difference-between-uri-escape-and-cgi-escape)". I've added integration test for the file name with spaces and removed the second encoding. 

@creepycheese Could you please review the PR and make sure that non uri encoded path works for you after removing double encoding? 

Thanks,

vishrutshah:
Respective code changes in azure-core gem is at [PR #3](https://github.com/Azure/azure-ruby-asm-core/pull/3).

@veronicagg Could you please review and let me know your feedback.

service_management/azure/test/integration/blob/create_page_blob_test.rb
```diff
@@ -80,4 +80,28 @@
...
+      subject.delete_container container_name
+    end
+
+    it 'should create a block blob with spaces in name' do
+      blob_name = 'blob with spaces'
+      blob = subject.create_block_blob container_name, blob_name, 'content'
+      blob.name.must_equal blob_name
+    end
+
+    it 'should create block blob with complex in name' do
```
veronicagg: is this validating something additional to "should create a block blob with spaces in name" and "creates page blob with non uri encoded path"?

vishrutshah: The second test validates the blob name with non uri encoded characters like `"with фбаф.txt"` to see that the those get persisted in the blob name on azure and can be passed on through the service.

veronicagg: oh, I see, thanks!

veronicagg:
@vishrutshah Looks good! thanks.

vishrutshah:
Thanks for the review!

service_management/azure/test/integration/blob/create_page_blob_test.rb
```diff
@@ -28,8 +28,8 @@
       subject.create_container container_name
     }
 
-    it 'creates a page blob' do
-      blob = subject.create_page_blob container_name, blob_name, length
+    it 'creates page blob with non uri encoded path' do
+      blob = subject.create_page_blob container_name, 'фбаф.jpg', length
       blob.name.must_equal blob_name
```
yaxia: Shouldn't it be: `blob.name.must_equal 'фбаф.jpg'` ?

vishrutshah: @yaxia Yest it must be. Currently these tests aren't recorded and running in the travis-ci but i verified that this is wrong.

PR #366, opened 2016-04-15
Removing core, to move it as a separate gem

veronicagg:
Re-publishing PR to remove core after addressing mode changes in previous PR, and alphabetical ordering in gemspec.


vishrutshah:
@veronicagg changes looks good to me!

Issue #367, opened 2016-04-15
v0.7.3: `initialize': Signing key must be provided (ArgumentError)


AbelHu:
Below code works with v0.7.0 but does not work with v0.7.3. Please see error in below logs.

Code.
require 'azure'
client = Azure.client(storage_account_name: storage_name, storage_access_key: storage_key)
client.storage_blob_host = "https://#{storage_name}.blob.core.windows.net"
blob = client.blobs
or
client.storage_table_host = "https://#{storage_name}.table.core.windows.net"
table = client.tables

Logs:
/home/xxx/.rvm/gems/ruby-2.1.3@test/gems/azure-0.7.3/lib/azure/core/auth/signer.rb:32:in `initialize': Signing key must be provided (ArgumentError)
        from /home/xxx/.rvm/gems/ruby-2.1.3@test/gems/azure-0.7.3/lib/azure/core/auth/shared_key.rb:33:in `initialize'
        from /home/xxx/.rvm/gems/ruby-2.1.3@test/gems/azure-0.7.3/lib/azure/blob/blob_service.rb:26:in `new'
        from /home/xxx/.rvm/gems/ruby-2.1.3@test/gems/azure-0.7.3/lib/azure/blob/blob_service.rb:26:in `initialize'
        from /home/xxx/.rvm/gems/ruby-2.1.3@test/gems/azure-0.7.3/lib/azure/client_services.rb:23:in `new'
        from /home/xxx/.rvm/gems/ruby-2.1.3@test/gems/azure-0.7.3/lib/azure/client_services.rb:23:in `blobs'
        from ./main.rb:21:in `<main>'

AbelHu:
Above code works with v0.7.1, too.

vishrutshah:
Hi @AbelHu , we have fixed the problem and released a new version of azure gem v0.7.4. 

Please use [Azure Gem v0.7.4](https://rubygems.org/gems/azure/versions/0.7.4) and let us know your feedback. 

Thanks,

PR #368, opened 2016-04-15
Updating how default_client sets options 

veronicagg:
Addressing issue #367.

### DIFF
--- a/service_management/azure/lib/azure/client_services.rb
+++ b/service_management/azure/lib/azure/client_services.rb
@@ -92,7 +92,8 @@ module Azure
     private
 
     def default_client(opts)
-      {client: Azure.client(opts || {})}
+      !opts.empty? ? {client: Azure.client(opts || {})} : {client: self}
     end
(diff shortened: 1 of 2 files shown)


service_management/azure/lib/azure/client_services.rb
```diff
@@ -92,7 +92,8 @@ def base_management(options = {})
     private
 
     def default_client(opts)
-      {client: Azure.client(opts || {})}
+      !opts.empty? ? {client: Azure.client(opts || {})} : {client: self}
```
vishrutshah: As you mentioned offline that we can rewrite this as ` !opts.empty? ? {client: Azure.client(opts)} : {client: self}`

@veronicagg Other things looks great.

veronicagg: @vishrutshah updated. Thanks for the review!

vishrutshah:
@veronicagg LGTM!

PR #386, opened 2016-05-05
Removing ARM code, as we're placing it in a new repo

veronicagg:
Removing ARM related files, as we're splitting it into a new repo [https://github.com/Azure/azure-sdk-ruby/pull/1](url) . Updating README and travis file.


README.md
```diff
@@ -47,46 +29,9 @@ Additional info on Azure deployment models [https://azure.microsoft.com/en-us/do
```
vishrutshah: As per our gemspec, azure supports 1.9.3 and above. `s.required_ruby_version = '>= 1.9.3'` but i am not sure we want to take the support out yet.

vishrutshah:
@veronicagg  changes looks good!! Thanks,

README.md
veronicagg: you're right, I'll leave it as it now as I'm not changing it here, and we can decide on this later. thanks!

PR #423, opened 2016-08-16
[ARM] Replace data from recordings even when they are encoded

vishrutshah:


azurecla:

TTYL, AZPRBOT;

veronicagg:
@vishrutshah LGTM. Was this data not encoded in any case before, so we didn't see it?

vishrutshah:
@veronicagg Before that i haven't used those keys that is not the same as the after encoding. I noticed it that [both are being encoded](https://github.com/Azure/autorest/blob/45ddc1a84c58f5732a2f77418079c9439f39e892/src/client/Ruby/ms-rest-azure/lib/ms_rest_azure/credentials/application_token_provider.rb#L100) so that we should also cover that case as well.

vishrutshah:
Thanks for the review!!

PR #425, opened 2016-08-17
Adding tests for queues when using SAS

veronicagg:
Realized I had this test pending for review, they'll need an azure-core release with https://github.com/Azure/azure-ruby-asm-core/commit/2e5bc94ef4db3ea2a0fa910a7008f09472b63f28 to work, creating the PR anyway so after azure-core is released with the fix, we can validate this one and merge.
This integration test correspond to https://github.com/Azure/azure-ruby-asm-core/pull/10 fixed in azure-core


veronicagg:
@vishrutshah Please take a look when you get a chance, thanks!

test/vcr/vcr_cassettes/service_bus/test_0001_should_be_able_to_send_a_message_with_SAS_to_a_queue.yml
```diff
@@ -0,0 +1,169 @@
+---
```
vishrutshah: I was expecting key 'SB_SAS_KEY' in this file but i don't see it or i missed it somewhere. could you please double check?

veronicagg: the key is not sent, it's used to sign a string and produce a token, going in the headers "Authorization", which is deleted from the recording.

vishrutshah: Cool. Makes sense. Thanks!

---
In reply to: [76116475](https://github.com/Azure/azure-sdk-for-ruby/pull/425#discussion_r76116475) [](ancestors = 76116475)

lib/azure/service_bus/auth/shared_access_signer.rb
```diff
@@ -6,15 +6,15 @@ module Auth
       class SharedAccessSigner < Azure::Core::Auth::Signer
 
         # The number of seconds from the time of signature that the SAS token will expire
-        attr_accessor :expiry_offset, :key_name
+        attr_accessor :expiry_offset, :key_name, :access_key
```
vishrutshah: >access_key [](start = 50, length = 10)

Don't you already have an access_key attribute. I'd think we don't need to redefine it again here.

vishrutshah:
Looks like you missed the recording file in the last commit or we don't need it?

vishrutshah:
@veronicagg LGTM!

veronicagg:
@vishrutshah thanks for the review!

PR #439, opened 2016-08-26
Be inclusive of which files to bundle with gems

vishrutshah:
This change is about being inclusive on which files to be bundled with gems instead of rejecting files do not required from git ls-files -z

This way we will not unintentional bundle files like Gemfile, *.gemspec, Rakefile etc... like we already did that until recently.

### DIFF
--- a/azure_mgmt_authorization/azure_mgmt_authorization.gemspec
+++ b/azure_mgmt_authorization/azure_mgmt_authorization.gemspec
@@ -18,7 +18,7 @@ Gem::Specification.new do |spec|
   spec.homepage      = 'https://aka.ms/azure-sdk-for-ruby'
   spec.license       = 'MIT'
 
-  spec.files         = `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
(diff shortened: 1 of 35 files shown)


azurecla:

TTYL, AZPRBOT;

azure_mgmt_authorization/azure_mgmt_authorization.gemspec
```diff
@@ -18,7 +18,7 @@ Gem::Specification.new do |spec|
   spec.homepage      = 'https://aka.ms/azure-sdk-for-ruby'
   spec.license       = 'MIT'
 
-  spec.files         = `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
+  spec.files         = Dir["LICENSE.txt", "lib/**/*"]
```
veronicagg: do we want to exclude Rakefile and gemspec file?

vishrutshah: Yup. I believe that they are not required.

veronicagg:
@vishrutshah LTMG!

PR #443, opened 2016-08-31
Adding require for azure/core and autoload Auth::SharedAccessSigner

veronicagg:
Adding require for azure/core and autoload Auth::SharedAccessSigner, so they don't need to be required from the caller. This needs also a change in azure-core, as azure-core is currently not autoloading auth files, see PR https://github.com/Azure/azure-ruby-asm-core/pull/25.

### DIFF
--- a/lib/azure.rb
+++ b/lib/azure.rb
@@ -76,9 +76,11 @@ module Azure
     autoload :SqlFilter,                'azure/service_bus/sql_filter'
     autoload :TrueFilter,               'azure/service_bus/true_filter'
     autoload :CorrelationFilter,        'azure/service_bus/correlation_filter'
+
     module Auth
       autoload :SharedAccessSigner,     'azure/service_bus/auth/shared_access_signer'
     end
+
   end
 
   module SqlDatabaseManagement
--- a/lib/azure.rb
+++ b/lib/azure.rb
@@ -22,6 +22,7 @@ require 'rexml/document'
 require 'addressable/uri'
 require 'faraday'
 require 'faraday_middleware'
+require 'azure/core'
 
 module Azure
   autoload :Client,                           'azure/client'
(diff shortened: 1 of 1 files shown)


lib/azure.rb
```diff
@@ -22,6 +22,7 @@
 require 'addressable/uri'
 require 'faraday'
 require 'faraday_middleware'
+require 'azure/core'
```
vishrutshah: With this change, Should we be removing all the unnecessary explicit requires of individual azure/core files?
Example: [require 'azure/core/signed_service'](https://github.com/Azure/azure-sdk-for-ruby/blob/asm/lib/azure/service_bus/service_bus_service.rb#L15)

lib/azure.rb
```diff
@@ -75,6 +76,9 @@ module ServiceBus
     autoload :SqlFilter,                'azure/service_bus/sql_filter'
     autoload :TrueFilter,               'azure/service_bus/true_filter'
     autoload :CorrelationFilter,        'azure/service_bus/correlation_filter'
+    module Auth
```
vishrutshah: minor: I'd leave an empty lines before `module `and after `end` for consistency

vishrutshah:
@veronicagg All the changes looks good. just let a comment to think about. Thanks!

lib/azure.rb
veronicagg: @vishrutshah Per our offline conversations, I prefer not to remove 'require' lines in this change, so we don't introduce a hard dependency on a new release of azure-core (that would autoload those classes).

lib/azure.rb
veronicagg: fixed.

lib/azure.rb
vishrutshah: Sounds good!

PR #469, opened 2016-09-21
[In Review] Adding specs & recordings for azure_mgmt_graph

vishrutshah:
This changes adds sample tests and recording on how to use Azure Graph services. There is a bug currently in release of [azure_mgmt_graph ](https://rubygems.org/gems/azure_mgmt_graph/versions/0.6.0) 0.6.0 with [ms_rest ](https://rubygems.org/gems/ms_rest/versions/0.5.0) 0.5.0 that prevent paging operation in graph. 

The fix for paging operation is currently in review @ https://github.com/Azure/autorest/pull/1443. This changes will be incorporated once new client runtime is released.


vishrutshah:
@veronicagg & @sarangan12 Feel free to review this PR when you get a chance. Thanks!

veronicagg:
@vishrutshah do we need any special tenet id, suscription, client id, secret, for this cases if we need to re-record them? Adding some simple tests cases cases seems fine, though having a sample and autorest test cases is probably what we want going forward. Changes look good to me. Thanks!

vishrutshah:
@veronicagg I don't think we need special tenet id, subscription but i guess if you are referring that as we were only able to do this test on msdn account then yes as i think it allows us more permission control. 

Regarding the tests cases in generator, Is that the one we covered in https://github.com/Azure/autorest/pull/1443 or you have specific case in mind?

PR #472, opened 2016-09-21
Build gems for verification in travis

vishrutshah:
As we had issues couple of time at the moment of releasing sdks they stops building due to various reasons, I am just including gem building process in the travis to avoid these surprises.


vishrutshah:
@veronicagg & @sarangan12 Feel free to review this PR when you get a chance. Thanks!

PR #477, opened 2016-09-23
Checking azure management sdks compatibility with JRuby

sarangan12:
Checking azure management sdks compatibility with JRuby

Travis Job Link: https://travis-ci.org/sarangan12/azure-sdk-for-ruby, https://travis-ci.org/sarangan12/azure-sdk-for-ruby/jobs/162281404

@veronicagg @vishrutshah Please review.


.travis.yml
```diff
@@ -5,6 +5,7 @@ rvm:
 - 2.1.5
 - 2.2.0
 - 2.3.0
+- jruby-9.1.5.0
```
vishrutshah: Just curious on the decision on the version of jruby. Why not `jruby-head` ?

sarangan12: JRuby seems to be going a lot of changes through releases. So, I thought it would be safer for us to control the versions on our end and test everytime we have a major JRuby release. Also, noticed that we are using multiple versions of ruby. So, eventually we might come to a point where we might need to support multiple versions of jruby. 

But, if you prefer -head I can modify that too.

vishrutshah: I agree with your thoughts but today we don't say we support only JRuby x, y or z. Based on the investigation on JRuby you mentioned that "_incompatibility between standard ruby and jruby arises only if the ruby gems use native extensions_". 

With that regards, my thoughts were 

1. as today we don't have any native dependencies we are compatible with JRuby and any future changes in JRuby should not break us either. 
2. checks with head would help us detecting before hand rather than at the time of release of our sdk only. 

With these thoughts in mind, I am more inclined for us to be not bounded by the JRuby versions & keep testing the `jruby-head`

@sarangan12 & @veronicagg what do you guys think?

veronicagg: @vishrutshah I think your suggestion is reasonable, we can try using `jruby-head`, so we know that we work on the latest. If we start seeing problems, we can revisit. thanks!

sarangan12: @vishrutshah @veronicagg I tried adding jruby-head and that does not work at present due to different issue. It can be seen here: https://github.com/travis-ci/travis-ci/issues/6085

veronicagg: @ sarangan12 Thanks for trying this out and finding this!

Issue #490, opened 2016-10-12
Azure gem has hard dependency on JSON 1.8 and fails with newer ruby versions


toddysm:
Azure gem is failing with newer Ruby versions (2.3 and 2.4) that bundle JSON 2.x gem by default because the dependency on JSON 1.8 gem

vishrutshah:
@toddysm Thanks for reporting an issue with the latest version ruby due to JSON dependency. We are looking into it and we'll get back to you soon. 

Thanks for your time.

sarangan12:
Fixes checked in.

toddysm:
Confirmed it is working for us now.

Issue #482, opened 2016-10-04
Relax mime-types gem pinning


dLobatog:
Hi devs,

I was looking at the latest gemspec and saw that mime-types is pinned to be < 3.0. Is there any reason for it to not be 4.0?

https://github.com/Azure/azure-sdk-for-ruby/blob/azure-v0.7.6/azure.gemspec

The reason I'm asking is that other programs that use this gem might have a dependency on newer mime-types gems. See https://github.com/theforeman/foreman-packaging/pull/1368 for an example.

I wanted to make a pull request myself fixing the dep and running tests, but it seems like the structure of the repository changes when you make a release (and that's when you do the gem pinnings). Anything I could do to fix this feel free to ask.

sarangan12:
Fixes checked in . Closing this issue

PR #518, opened 2016-10-24
Remove Json dependency and modify mime-types dependency

sarangan12:
1. https://github.com/Azure/azure-sdk-for-ruby/issues/482
2. https://github.com/Azure/azure-sdk-for-ruby/issues/490 

@vishrutshah @veronicagg Can you review this PR?

### DIFF
--- a/azure.gemspec
+++ b/azure.gemspec
@@ -35,8 +35,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency('azure-core',              '~> 0.1')
   s.add_runtime_dependency('faraday',                 '~> 0.9')
   s.add_runtime_dependency('faraday_middleware',      '~> 0.10')
-  s.add_runtime_dependency('json',                    '~> 1.8')
-  s.add_runtime_dependency('mime-types',              ['>= 1', '< 3.0'])  # vagrant-share and other stuff relies on 1
+  s.add_runtime_dependency('mime-types',              ['>= 1', '< 4.0'])  # vagrant-share and other stuff relies on 1
   s.add_runtime_dependency('nokogiri',                '~> 1.6')
   s.add_runtime_dependency('systemu',                 '~> 2.6')
   s.add_runtime_dependency('thor',                    '~> 0.19')
@@ -50,3 +49,4 @@ Gem::Specification.new do |s|
   s.add_development_dependency('yard',                '~> 0.8')
   s.add_development_dependency('vcr',                 '~> 3.0')
 end
+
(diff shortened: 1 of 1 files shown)


azure.gemspec
```diff
@@ -50,3 +49,4 @@ Gem::Specification.new do |s|
   s.add_development_dependency('yard',                '~> 0.8')
   s.add_development_dependency('vcr',                 '~> 3.0')
 end
+
```
vishrutshah: I think we left our conversation half-way (From: https://github.com/Azure-Samples/compute-ruby-manage-vm/pull/3) regarding this one. I believe that we should follow coding standards of `Rails` & `Ruby` by only having `CRLF` character at the end of line instead a new line because `rubocop` will consider this as offense. 

What are your thoughts?

sarangan12: how would we do that? at present, the only way is to use rubymine.  i am still waiting for mine.

azure.gemspec
```diff
@@ -35,8 +35,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency('azure-core',              '~> 0.1')
   s.add_runtime_dependency('faraday',                 '~> 0.9')
   s.add_runtime_dependency('faraday_middleware',      '~> 0.10')
-  s.add_runtime_dependency('json',                    '~> 1.8')
-  s.add_runtime_dependency('mime-types',              ['>= 1', '< 3.0'])  # vagrant-share and other stuff relies on 1
+  s.add_runtime_dependency('mime-types',              ['>= 1', '< 4.0'])  # vagrant-share and other stuff relies on 1
```
sarangan12: removing 4.0 will result in a warning as open ended dependency. And the latest version is 3.1

vishrutshah: Should we use then `~>3.0` ?

veronicagg: If we used ~>3.0 wouldn't we be not allowing versions >=1, which we use to allow? If we work well with versions between 1 and 3+, then the above seems reasonable to me.

azure.gemspec
sarangan12: Using that git property is risky. It converts CLRF and LF vice versa freely and do not know the consequences. So, for now, I am Ok with adding an empty line or removing it altogether. (I am Ok with the red marker in PRs).

sarangan12:
@veronicagg @vishrutshah Your comments have been addressed. CI has been fixed. 

@veronicagg I have tested from Ruby 2.0.0 and it uses only json 1.8. So, There is no worries there.

azure.gemspec
vishrutshah: @sarangan12 & @veronicagg Did you guys get chance to talk about what were the breaking changes between 2.0 and 3.0 of mime-type gem and whether are we wrapping those apis or exposing those through our gem?

Issue #521, opened 2016-10-26
Make vcr cassette request matching smarter


vishrutshah:
As of today when we generate the new ARM SDKs from new swagger API version, our recordings breaks. Following are the top two reasons why:
1. VCR Cassette recording matches entire url
2. Most of the API versions are backward compatibles but have different api version 

Make it ignore the query parameters as they include the api_version parameter and other parameters that should be okay to ignore while playback.

PR #524, opened 2016-10-27
Ignore query parameters on recording playback

vishrutshah:
#521

Reference: https://www.relishapp.com/vcr/vcr/v/1-7-1/docs/cassettes/request-matching#match-on-host-and-path-(to-ignore-query-params)

With this fix in the recordings pattern matching, we'd not need to re-record recordings in case the query parameters changes (Ex. api_version) in the new swagger.

### DIFF
--- a/vcr_helper.rb
+++ b/vcr_helper.rb
@@ -16,7 +16,7 @@ if !ENV['INTEG_RECORDED'].nil? && ENV['INTEG_RECORDED'] == 'true'
 end
 
 VCR.configure do |config|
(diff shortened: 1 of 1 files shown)


vishrutshah:
@veronicagg & @sarangan12 please review it when you get a chance. Thanks!

vishrutshah:
@veronicagg I've made changes as per the review. Please take a look when you get a chance. Thanks!

vishrutshah:
Thanks for the review!!

PR #527, opened 2016-10-28
Use composite swagger for Azure Recovery backup service

vishrutshah:
Using composite swagger spec from https://github.com/Azure/azure-rest-api-specs/tree/master/arm-recoveryservicesbackup


vishrutshah:
@sarangan12 & @veronicagg Please review as you get a chance. Thanks!

vishrutshah:
Thanks for the review!!

Issue #528, opened 2016-10-28
azure_mgmt_sql should be generated from composite swagger


veronicagg:
azure_mgmt_sql should be generated using composite swagger at https://raw.githubusercontent.com/Azure/azure-rest-api-specs/master/arm-sql/compositeSql.json
(we should be always generating from composite swagger when available)

PR #530, opened 2016-10-28
Use composite swagger for Azure Sql service

vishrutshah:
https://github.com/Azure/azure-sdk-for-ruby/issues/528


vishrutshah:
Thanks for the review!!

Issue #517, opened 2016-10-24
Can we customize user-agent for consumer gems


vishrutshah:
Investigate
- Are we exposing the user-agent customization
- How easy is it to add new user-agent string with each call
- How easy is it to add new user-agent for consuming gems' all calls

vishrutshah:
Here are the answers to the above questions and idea for implementation: 

1. Are we exposing the user-agent customization?
Yes but not streamlined. [HttpOperationRequest](https://github.com/Azure/azure-sdk-for-ruby/blob/master/runtime/ms_rest/lib/ms_rest/http_operation_request.rb) contains an attribute names as `user_agent_extended` that can take array of string to be appended as `User-Agent` header for requests. 

2. How easy is it to add new user-agent string with each call?
It can be added through each method call from operation as `custom_headers ` parameter hash and adding a key of `User-Agent` that will be merged in runtime before making a request in the method call [run_promise](https://github.com/Azure/azure-sdk-for-ruby/blob/master/runtime/ms_rest/lib/ms_rest/http_operation_request.rb#L80). This doesn't take advantage of `user_agent_extended` parameters. 

3. How easy is it to add new user-agent for consuming gems' all calls?
Every consumer needs to take advantage of the `custom_headers` parameter for each calls to our sdk. It's pain to add header for each call unless customers wants to track how each requests.

**Requirements**
- We should have streamlined option for setting values once per client to be used in each call. 
- Values set on the client should be overridable by each call's `custom_headers` option

**Implementation Proposal** 
Adding `user_agent_extended` on the [ServiceClient](https://github.com/Azure/azure-sdk-for-ruby/blob/master/runtime/ms_rest/lib/ms_rest/service_client.rb) will enable each client of the azure_mgmt_* sdks to optionally set the values to be sent to the Azure Services in the part of `User-Agent` headers. 

The sample code would look like below:
```
  require 'azure_mgmt_compute'
  compute_client = ComputeManagementClient.new(@credentials)
  compute_client.subscription_id = 'xxxxxx-xxxxxx-xxxxxx-xxxxx'
  compute_client.user_agent_extended.push('azure_mgmt_compute')
```

Here is the request header would look like:
```
    headers:
      User-Agent:
      - Azure-SDK-For-Ruby/<MS_REST_VERSION>/azure_mgmt_compute
      Content-Type:
      - application/json; charset=utf-8
      accept-language:
      - en-US
      x-ms-client-request-id:
[...]

PR #543, opened 2016-11-03
Extend user_agent_extended array to add telemetry information

vishrutshah:
This is part of the feature #517. We've exposed `user_agent_extended` array on the [ServiceClient ](https://github.com/Azure/azure-sdk-for-ruby/blob/master/runtime/ms_rest/lib/ms_rest/service_client.rb#L9) of `ms_rest` to be used for appending additional telemetry information to be sent via `User-Agent` header. 

**User-Agent Header Before** from [AzureServiceClient](https://github.com/Azure/azure-sdk-for-ruby/blob/master/runtime/ms_rest_azure/lib/ms_rest_azure/azure_service_client.rb#L9)
```
Azure-SDK-For-Ruby/0.6.0/
```

**User-Agent Header After** from [AzureServiceClient](https://github.com/Azure/azure-sdk-for-ruby/blob/master/runtime/ms_rest_azure/lib/ms_rest_azure/azure_service_client.rb#L9)
```
Ruby/2.2.4 (x86_64-darwin15) ms_rest/0.6.0 ms_rest_azure/0.6.0 Azure-SDK-For-Ruby
```

After this change, we can also leverage [add_user_agent_information](https://github.com/Azure/azure-sdk-for-ruby/pull/543/files#diff-3970cf987aba9f07068141e4e141d0e7R71) method to add more information about azure_mgmt_* sdks from each of the management clients & consumer of azure_mgmt_* sdks can also append more information as well for us to have rich telemetry.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ script:
 - bundle exec rake arm:build
 - unset BUNDLE_GEMFILE
 - cd $TRAVIS_BUILD_DIR/runtime/ms_rest && bundle install --gemfile=Gemfile && bundle exec rspec
-- unset BUNDLE_GEMFILE
 - cd $TRAVIS_BUILD_DIR/runtime/ms_rest_azure && bundle install --gemfile=Gemfile && bundle exec rspec
 deploy:
   provider: script
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,12 +6,13 @@ rvm:
 - 2.2.0
 - 2.3.0
 script:
-- cd $TRAVIS_BUILD_DIR/runtime/ms_rest && bundle install && bundle exec rspec
-- cd $TRAVIS_BUILD_DIR/runtime/ms_rest_azure && bundle install && bundle exec rspec
-- cd $TRAVIS_BUILD_DIR
-- bundle install
+- gem install bundler
 - if [ "$INTEG_RECORDED" == "true" ] ; then bundle exec rake arm:spec ; fi
 - bundle exec rake arm:build
+- unset BUNDLE_GEMFILE
+- cd $TRAVIS_BUILD_DIR/runtime/ms_rest && bundle install --gemfile=Gemfile && bundle exec rspec
+- unset BUNDLE_GEMFILE
+- cd $TRAVIS_BUILD_DIR/runtime/ms_rest_azure && bundle install --gemfile=Gemfile && bundle exec rspec
 deploy:
   provider: script
   script: ./scripts/release.sh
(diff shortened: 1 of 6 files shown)


vishrutshah:
@veronicagg & @sarangan12 Feel free to take a look when you get a chance. Thanks!

.travis.yml
```diff
@@ -11,7 +11,6 @@ script:
 - bundle exec rake arm:build
 - unset BUNDLE_GEMFILE
 - cd $TRAVIS_BUILD_DIR/runtime/ms_rest && bundle install --gemfile=Gemfile && bundle exec rspec
-- unset BUNDLE_GEMFILE
```
vishrutshah: https://docs.travis-ci.com/user/languages/ruby/#%24BUNDLE_GEMFILE-environment-variable

runtime/ms_rest_azure/lib/ms_rest_azure/azure_service_client.rb
```diff
@@ -19,6 +19,8 @@ def initialize(credentials, options = nil)
       @request_headers =  {
           'Content-Type' => 'application/json;charset=utf-8' # This is the current default for Azure services, and content-type supported by Autorest
       }
+      add_user_agent_information("ms_rest_azure/#{MsRestAzure::VERSION}")
+      add_user_agent_information("Azure-SDK-For-Ruby")
```
veronicagg: are we including "Azure-SDK-For-Ruby" string for compatibility with already collected data? since it's getting logged together with ms_rest_azure info, does it give us additional benefits?

vishrutshah: Yes, we are including "Azure-SDK-For-Ruby" string to make sure current telemetry system works as well as that is the defined string for each language "Azure-SDK-For-{LANG}" for telemetry collection. So the purpose it two-fold 1. Keep existing telemetry alive and 2. That is the current way to distinguish between languages. 

Regarding keeping "Azure-SDK-For-Ruby" on ms_rest_azure i am not sure what could be the best decision because that actually represents the SDK and should go into each management client but on the other hand keeping it in each client seems overhead. As we are the one going to interpret the data, I am thinking it'd be fine to put that string in ms_rest_azure. What do you think?

.travis.yml
vishrutshah: @veronicagg I verified that there is not need to keep two unset as running bundle install from script section of the travis doesn't set or affect that environment variable.

runtime/ms_rest_azure/lib/ms_rest_azure/azure_service_client.rb
veronicagg: Sounds good, the string looks redundant but it's needed for data collection/analysis and consistency, so I'm fine having it in the runtime. Thanks!

vishrutshah:
@veronicagg Thanks for the review!!

PR #595, opened 2016-11-16
Regen azure_mgmt_service_bus using AutoRest.1.0.0-Nightly20161114

vishrutshah:
Part of the feature work https://github.com/Azure/azure-sdk-for-ruby/issues/517


vishrutshah:
Thanks for the review!!

PR #599, opened 2016-11-22
add model to_json, from_json and to_s mixin

devigned:
I found it painful to debug output from the lib since the models don't provide much information when called via to_s. This pull request adds to and from json as well as a more robust to_s implementation which includes the json representation of the model.

### DIFF
--- a/runtime/ms_rest/lib/ms_rest/jsonable.rb
+++ b/runtime/ms_rest/lib/ms_rest/jsonable.rb
@@ -1,0 +1,35 @@
+# encoding: utf-8
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+
+module MsRest
+  # Mixin to provide simple serialization / deserialization in AutoRest generated model classes
+  module JSONable
+    include MsRest::Serialization
+
+    #
+    # Serialize the object to JSON
+    # @return [String] JSON serialized version of the object
+    #
+    def to_json
+      serialize(self.class.mapper, self)
+    end
+
+    #
+    # Deserialize the object from JSON
+    # @param json [String] JSON string representation of the object
+    # @return [JSONable] object built from json input
+    #
+    def from_json(json)
+      deserialize(self.class.mapper, json)
+    end
+
+    #
+    # String representation of the object
+    # @return [String]
+    #
+    def to_s
+      "#{super} #{to_json.to_s}"
+    end
+  end
+end
--- a/runtime/ms_rest/lib/ms_rest/serialization.rb
+++ b/runtime/ms_rest/lib/ms_rest/serialization.rb
@@ -15,7 +15,7 @@ module MsRest
     # @param response_body [Hash] Ruby Hash object to deserialize.
     # @param object_name [String] Name of the deserialized object.
     #
-    def deserialize(mapper, response_body, object_name)
+    def deserialize(mapper, response_body, object_name = nil)
       build_serializer.deserialize(mapper, response_body, object_name)
     end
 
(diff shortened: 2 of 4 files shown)


salameer:
Adding captain Sarangan to review, our current and only Ruby resident :) @sarangan12

devigned:
@sarangan12 thoughts?

runtime/ms_rest/lib/ms_rest/jsonable.rb
```diff
@@ -0,0 +1,22 @@
+# encoding: utf-8
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+
+module MsRest
+  # Mixin to provide simple serialization / deserialization in AutoRest generated model classes
+  module JSONable
+    include MsRest::Serialization
+
+    def to_json
```
vishrutshah: Minor: Although added methods are self explanatory, It'd be awesome to have a minor comments on each added methods to_json, from_json and to_s for documentation perspective. What do you think?

runtime/ms_rest/lib/ms_rest/serialization.rb
```diff
@@ -391,7 +391,14 @@ def serialize_sequence_type(mapper, object, object_name)
       # @param model_name [String] Name of the model to retrieve.
       #
       def get_model(model_name)
-        Object.const_get(@context.class.to_s.split('::')[0...-1].join('::') + "::Models::#{model_name}")
+        consts = @context.class.to_s.split('::')
+        if consts.any?{ |const| const == 'Models' }
```
vishrutshah: The way this logic has been updated, it seems like there was a bug in discovering the model class. Just curious as it seems like you ran into it. :)

runtime/ms_rest/lib/ms_rest/jsonable.rb
devigned: Good point. I'll update the pull request with documentation. Thank you for catching this.

runtime/ms_rest/lib/ms_rest/serialization.rb
devigned: Not so much a bug, but another way of including the serialization logic. Honestly, the serialization logic shouldn't need an anchor to the service class. 

JSONable will be decorated onto the model, rather than the @client, and as such, needs to understand the context so it can create a new version of the deserialized object.

Perhaps, this logic should be decorated onto the model and not known by the service class. I was thinking about doing this larger refactor, but didn't feel like it was my place to make sweeping changes like that.

Does this make sense?

veronicagg: Interesting, do you mind opening an issue to track this potential refactor? We can evaluate it and decide if we want to pursue it. Thanks!

devigned: Uhm, sure. Basically, you'd change from:
```ruby
      # Serialize Request
      request_mapper = VirtualMachine.mapper()
      request_content = @client.serialize(request_mapper,  parameters, 'parameters')
      request_content = request_content != nil ? JSON.generate(request_content, quirks_mode: true) : nil
      # ....
      result_mapper = VirtualMachine.mapper()
      result.body = @client.deserialize(result_mapper, parsed_response, 'result.body')
```
To:
```ruby
      # Serialize Request
      request_content = VirtualMachine.serialize(parameters, name: 'parameters', quirks_mode: true)
      # ....
      result.body = VirtualMachine.deserialize(parsed_response, name: 'result.body')
```
The client really doesn't need to be involved. It's coupled to serialization only because the deserializer needs to know how to retrieve the model type. If the serialization logic is decorated on the model, this becomes trivial, insatiate a `self.class`.

devigned:
I should get to these updates this evening.

devigned:
@vishrutshah I believe your feedback should be handled now.

Issue #566, opened 2016-11-10
VirtualMachine on poweroff/poweron/start has to response. You just get Nil


jkwon2013:
Previous history on the issue from Puppet's Trello bug board:
David Justice: Take a look at this code: https://github.com/Azure/vagrant-azure/blob/v2.0/lib/vagrant-azure/action/read_state.rb

cc:
@gregohardy

sarangan12:
The [Power Off](https://github.com/Azure/azure-rest-api-specs/blob/master/arm-compute/2016-03-30/swagger/compute.json#L1297), [Restart](https://github.com/Azure/azure-rest-api-specs/blob/master/arm-compute/2016-03-30/swagger/compute.json#L1334), [Start](https://github.com/Azure/azure-rest-api-specs/blob/master/arm-compute/2016-03-30/swagger/compute.json#L1371) are all long running operations and return nil. 

This seems to be a bug in the code. Because the similar long running operation create does not return nil. On looking into the azure-sdk-for-ruby code, you can find the following:
```
........
........
def start(resource_group_name, vm_name, custom_headers = nil)
  response = start_async(resource_group_name, vm_name, custom_headers).value!
  nil
end
........
........
def create_or_update(resource_group_name, vm_name, parameters, custom_headers = nil)
  response = create_or_update_async(resource_group_name, vm_name, parameters, custom_headers).value!
  response.body unless response.nil?
end
.........
.........
```

As you can see, the nil is intentionally returned and this causes the issue. I will work on the fix and update this issue soon.

sarangan12:
On further analysis, it seems in addition to the specified methods, the following methods are also affected by the same behavior. (Needs verification & confirmation)

1. delete
2. delete_async
3. deallocate
4. deallocate_async
5. generalize
6. power_off	
7. power_off_async
8. restart
9. restart_async
10. start
11. start_async
12. redeploy
13. redeploy_async
14. begin_delete
15. begin_deallocate
16. begin_power_off
17. begin_restart
18. begin_start
19. begin_redeploy

I will verify and confirm the same.

sarangan12:
@gregohardy The issue has been fixed now in the master branch and must be available in the next release.

PR #608, opened 2016-11-30
Added code to avoid returning nil values

sarangan12:
In the get_long_running_operation_result method, a custom_deserialization_block has been passed. The idea is to call this block once we receive the response. But, if you look at the specific line of code change update_state_from_azure_async_operation_header method does not even get the block as input. This method should be called only for standard deserialization i.e. when there is no custom block. Else, there will not be any use in passing the custom block. 

So, I have added a condition to check for the existence of custom block. @salameer @veronicagg @vishrutshah Please review. 

Related to https://github.com/Azure/azure-sdk-for-ruby/issues/566


runtime/ms_rest_azure/lib/ms_rest_azure/azure_service_client.rb
```diff
@@ -41,7 +41,7 @@ def get_long_running_operation_result(azure_response, custom_deserialization_blo
       if !AsyncOperationStatus.is_terminal_status(polling_state.status)
         task = Concurrent::TimerTask.new do
           begin
-            if !polling_state.azure_async_operation_header_link.nil?
+            if !polling_state.azure_async_operation_header_link.nil? && custom_deserialization_block.nil?
```
veronicagg: @sarangan12 can you describe a little more how this additional condition helps with not returning nil values?

veronicagg: @amarzavery here's a more specific question to you, could you give us some more context on the use of the operation header link versus the location header link and how we should be consuming them? Our code without this change is not using deserialization knowledge for the operation header, though this knowledge is used to deserialize the response when no operation header link exists and location header is present. I'd like to better understand if there's a reason for that, or just a bug in our code.

veronicagg: @sarangan12 With the proposed fix if we have operation header and no location header and customer_deserialization_block is not nil (and is not a 'put'), we return AzureOperationError 'Location header is missing...', is that the correct result for this case? Is that case possible? Without the fix, in that case we would go in the first "if" clause, and execute different code, would that give a different result?

sarangan12: @veronicagg 
1. If we send a custom deserialization block, 
2. the response does have a async operation header link 
3. No location header link 
4. And Not a Put Method

Yes. I agree that would be an issue. But is there a scenario where only one of the links is provided and the other one is not present?

sarangan12:
Note: Before the proposed changes: if a response has only location header with no deserialization block it would have taken a different route. Now, it is taking a different route. Since I do not see any significant change between them (except for deserialization), I believe it should be fine.

runtime/ms_rest_azure/spec/azure_service_client_spec.rb
```diff
@@ -37,7 +37,7 @@ module MsRestAzure
       azure_service_client.long_running_operation_retry_timeout = 0
 
       allow_any_instance_of(MsRestAzure::PollingState).to receive(:create_connection).and_return(nil)
-      allow(azure_service_client).to receive(:update_state_from_azure_async_operation_header) do |request, polling_state|
+      allow(azure_service_client).to receive(:update_state_without_custom_deserialization) do |request, polling_state|
```
veronicagg: with this change, how are we validating the intent of the test "should use async operation header when async_operation_header present" ? it looks like we're using the same method name in both sets of tests (for async operation header and for location header check.

runtime/ms_rest_azure/spec/azure_service_client_spec.rb
```diff
@@ -64,13 +64,13 @@ module MsRestAzure
       azure_service_client.long_running_operation_retry_timeout = 0
 
       allow_any_instance_of(MsRestAzure::PollingState).to receive(:create_connection).and_return(nil)
-      allow(azure_service_client).to receive(:update_state_from_location_header) do |request, polling_state|
+      allow(azure_service_client).to receive(:update_state_without_custom_deserialization) do |request, polling_state|
```
veronicagg: similar question here (as previous), the method here, does not differentiate aync_operation_header versus location_header, how are we validating the intent of the test now? and with the new logic are we still complying with the intent of the test?

runtime/ms_rest_azure/lib/ms_rest_azure/azure_service_client.rb
```diff
@@ -41,10 +41,12 @@ def get_long_running_operation_result(azure_response, custom_deserialization_blo
       if !AsyncOperationStatus.is_terminal_status(polling_state.status)
         task = Concurrent::TimerTask.new do
           begin
-            if !polling_state.azure_async_operation_header_link.nil?
-              update_state_from_azure_async_operation_header(polling_state.get_request(headers: request.headers, base_uri: request.base_uri, user_agent_extended: user_agent_extended), polling_state)
-            elsif !polling_state.location_header_link.nil?
-              update_state_from_location_header(polling_state.get_request(headers: request.headers, base_uri: request.base_uri, user_agent_extended: user_agent_extended), polling_state, custom_deserialization_block)
+            if !polling_state.azure_async_operation_header_link.nil? || !polling_state.location_header_link.nil?
```
vishrutshah: Does this condition means that now as long as you find either `azure_async_operation_header_link` or `location_header_link` with `custom_deserialization_block` you'll call `update_state_with_custom_deserialization`? Because if i understand correctly, we were only calling update_state_without_custom_deserialization (old name: update_state_from_azure_async_operation_header) only when we have `azure_async_operation_header_link` not when we find `location_header_link`?

sarangan12: Does this condition means that now as long as you find either azure_async_operation_header_link or location_header_link with custom_deserialization_block you'll call update_state_with_custom_deserialization?

Yes.

Because if i understand correctly, we were only calling update_state_without_custom_deserialization (old name: update_state_from_azure_async_operation_header) only when we have azure_async_operation_header_link not when we find location_header_link?

It does not matter. Since the polling_state.get_request method has an or condition. It picks up async_operation_header_link if it is present. Else, it picks up the location header link

sarangan12:
Creating a new PR

Issue #340, opened 2016-03-31
Resource Group is available on resources


devigned:
Resource group is not available as a property on the virtual machine. You can ascertain it from the URI. This is a pain as it will probably change in the future. An accessor would be better.

All resources should provide the resource group based on the uri of the resource.

veronicagg:
we could obtain the resource group name from the ID of the MsRestAzure::Resource::ID, and surface it to the properties.

sarangan12:
Closing the issue

PR #620, opened 2016-12-13
Add Resource Group to Resources

sarangan12:
Related to Azure/azure-sdk-for-ruby#340

@veronicagg @vishrutshah @salameer Please review.


runtime/ms_rest_azure/lib/ms_rest_azure/resource.rb
```diff
@@ -23,6 +23,9 @@ class Resource
     # @return [Hash{String => String}] the tags attached to resources (optional).
     attr_accessor :tags
 
+    # @return [String] the resource group of the resource.
+    attr_accessor :resource_group
```
veronicagg: should we call it resource_group_name? and update the description to indicate this is the name of the resource group?

runtime/ms_rest_azure/lib/ms_rest_azure/resource.rb
```diff
@@ -23,6 +23,15 @@ class Resource
     # @return [Hash{String => String}] the tags attached to resources (optional).
     attr_accessor :tags
 
+    # @return [String] the name of the resource group of the resource.
+    def resource_group
+      unless self.id.nil?
+        if self.id =~ /.+\/resourceGroups\/.+\/.+/
```
vishrutshah: I'd recommend below pattern as Group based matching, so that you don't have to split by `/` afterwards:

```
groups = self.id.match(/.*\/resourceGroups\/(.*)\/.*/)
groups.captures[0] if groups
```

runtime/ms_rest_azure/lib/ms_rest_azure/resource.rb
```diff
@@ -23,6 +23,15 @@ class Resource
     # @return [Hash{String => String}] the tags attached to resources (optional).
     attr_accessor :tags
 
+    # @return [String] the name of the resource group of the resource.
+    def resource_group
```
vishrutshah: One more optimization thought is as `resource_group` is a computed property of `Resource` class, we should only compute it once when we are setting up `id` not every time we access the `resource_group`.

Issue #659, opened 2017-01-19
Generate Models for SubResource


sarangan12:

sarangan12:
Merge completed to master branch

PR #660, opened 2017-01-19
Removed the SubResource - Associated Spec file. Modified a related test case

sarangan12:
Related to Azure/azure-sdk-for-ruby#659

@salameer @veronicagg @vishrutshah @devigned  Please review.


devigned:
@sarangan12 I thought the subresource was to still exist in msrest, but would be subclassed in the case of redefinition (with additional properties) within a generated client.

After thinking about this a bit more, I'm coming to terms with what subresource actually represents. I think the real goal of subresource is to recognize that a given class is "Identifiable", that is, it has an id property on it. This serves two goals (off the top of my head, perhaps more):
- Having SubResource lowers the number of generated classes
- Provides a common base rather than asking a given object if they respond_to :id.

The question is, are those features valuable. If so, then we should have some way to have a common class (or module) for id bearing objects.

management/azure_mgmt_compute/spec/virtual_machines_spec.rb
```diff
@@ -135,7 +135,7 @@
 
     result = @resource_helper.network_client.network_interfaces.list_all
     result.each do |network_interface|
-      expect(network_interface.virtual_machine).to be_an_instance_of(MsRestAzure::SubResource)
+      expect(network_interface.virtual_machine).to be_an_instance_of(Azure::ARM::Network::Models::SubResource)
```
vishrutshah: Just curious: Is there any reason that you need full namespace to access SubResource?

Issue #647, opened 2017-01-11
Azure Recovery Services Vaults Sku Attribute


bilal-naeem-confiz:
Hi @vishrutshah 

I am using the gem `azure_mgmt_recovery_services` to create a vault. [L131](https://github.com/Azure/azure-sdk-for-ruby/blob/master/management/azure_mgmt_recovery_services/lib/generated/azure_mgmt_recovery_services/vaults.rb#L131) mentions that a `Azure::ARM::RecoveryServices::Models::Vault` object is also required to be passed to the `create_or_update` operation. So I created a `Vault` object and passed it to the `create_or_update` call. My script is the following
```
require 'azure_mgmt_recovery_services'

subscription_id = '########-####-####-####-############'
provider = MsRestAzure::ApplicationTokenProvider.new('########-####-####-####-############',
                                                     '########-####-####-####-############',
                                                     '############################################'
)
credentials = MsRest::TokenCredentials.new(provider)

recovery_services = Azure::ARM::RecoveryServices::RecoveryServicesClient.new(credentials)
recovery_services.subscription_id = subscription_id

vault = Azure::ARM::RecoveryServices::Models::Vault.new
vault.name = 'TestVault'
vault.location = 'eastus'
vault.properties = Azure::ARM::RecoveryServices::Models::VaultProperties.new

puts recovery_services.vaults.create_or_update('TestRG-RV', 'TestVault', vault).inspect
```

However, running this script results in the following error
```
/var/lib/gems/2.3.0/gems/azure_mgmt_recovery_services-0.8.0/lib/generated/azure_mgmt_recovery_services/vaults.rb:205:in `block in create_or_update_async': { (MsRestAzure::AzureOperationError)
  "message": "MsRestAzure::AzureOperationError: InvalidRestApiParameter: properties parameter is invalid.\r\nPlease provide a valid properties",
  "request": {
    "base_uri": "https://management.azure.com",
    "path_template": "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.RecoveryServices/vaults/{vaultName}",
    "method": "put",
    "path_params": {
      "subscriptionId": "67f2116d-4ea2-4c6c-b20a-f92183dbe3cb",
      "resourceGroupName": "TestRG-RV",
      "vaultName": "TestVault"
    },
    "skip_encoding_path_params": null,
    "query_params": {
      "api-version": "2016-06-01"
    },
[...]

PR #663, opened 2017-01-30
Added SubResource/Resource Hierarchy

sarangan12:
Context for this PR:
In the Swagger Files, there will be a definition for "Resource" and "SubResource" (if they are used). These definitions have historically remained same across the services. 

Resource: name, id, type, location, tags
SubResource: id

Now, there are services such as datalakeanalyticslake, datalakeanalyticsservice, recoveryservicesbackup, etc have added additional properties to them (which is valid). For example, Sku has been added to Resources. 

How are we going to handle it? 

Option 1: We could generate the Resource and SubResource classes per service. This is the approach taken by node sdk. This has been discussed and rejected.

Option 2: We could keep the Resource and SubResource classes as such in the runtime. We could generate the similar classes per service only when they diverge from the standard definition. This is the approach taken by this PR. As an additional enhancement, we have established the relation between the SubResource and Resource classes themselves. 

Note: One another suggestion was to rename the class "SubResource" to "Identifiable". This is a valid suggestion. But, this needs additional considerations. Are we going to reflect the namechange in all SDKs? Are we going to use that Identifiable class as parent class for any model that has only the Id property? There may be other questions. I have made a call to not do that in this PR. I have opened a seperate issue #664

Ref #1: https://github.com/Azure/azure-sdk-for-ruby/issues/659
Ref #2: https://github.com/Azure/azure-sdk-for-ruby/issues/647 
Ref #3: Failure of nightly builds

@veronicagg @vishrutshah @salameer Please review.

### DIFF
--- a/runtime/ms_rest_azure/lib/ms_rest_azure/resource.rb
+++ b/runtime/ms_rest_azure/lib/ms_rest_azure/resource.rb
@@ -2,7 +2,6 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License. See License.txt in the project root for license information.
 
-require_relative 'sub_resource'
 
 module MsRestAzure
   #
--- a/runtime/ms_rest_azure/lib/ms_rest_azure/resource.rb
+++ b/runtime/ms_rest_azure/lib/ms_rest_azure/resource.rb
@@ -2,14 +2,13 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License. See License.txt in the project root for license information.
 
+require_relative 'sub_resource'
+
 module MsRestAzure
   #
   # Class which represents any Azure resource.
   #
-  class Resource
-
-    # @return [String] the id of the resource.
-    attr_accessor :id
+  class Resource < SubResource
 
     # @return [String] the name of the resource.
     attr_accessor :name
--- a/runtime/ms_rest_azure/lib/ms_rest_azure.rb
+++ b/runtime/ms_rest_azure/lib/ms_rest_azure.rb
@@ -12,9 +12,9 @@ require 'ms_rest_azure/azure_service_client.rb'
 require 'ms_rest_azure/cloud_error_data.rb'
 require 'ms_rest_azure/credentials/application_token_provider.rb'
 require 'ms_rest_azure/polling_state.rb'
+require 'ms_rest_azure/sub_resource.rb'
 require 'ms_rest_azure/resource.rb'
 require 'ms_rest_azure/serialization.rb'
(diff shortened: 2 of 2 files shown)


runtime/ms_rest_azure/lib/ms_rest_azure/resource.rb
```diff
@@ -2,14 +2,13 @@
...
+
 module MsRestAzure
   #
   # Class which represents any Azure resource.
   #
-  class Resource
-
-    # @return [String] the id of the resource.
-    attr_accessor :id
+  class Resource < SubResource
```
vishrutshah: Just curious why not use full namespace while inhering to avoid conflicts like `class Resource < MsRestAzure::SubResource`. That should also eliminate need for `require_relative 'sub_resource'`.

Moreover the name `Sub_Resource` inherited into `Resource` doesn't make sense as developer. I'd suggest renaming it as you mentioned that `SubResource` is not at all Azure concept so we better avoid explicitly referring it as SubResource.

sarangan12: This needs change in the order in another file. It has been done with the latest commit

sarangan12:
Adding @devigned

runtime/ms_rest_azure/lib/ms_rest_azure/resource.rb
sarangan12: Regarding renaming, I think it has some more items to discuss. Let me create a WI and add it to the backlog https://github.com/Azure/azure-sdk-for-ruby/issues/664

veronicagg:
@sarangan12 could you add some context to the issue related to this PR (and reference it)? for our reference later on it'd be good to describe what problem is being solved, what options were evaluated, what option was picked for implementation and why.

runtime/ms_rest_azure/lib/ms_rest_azure/resource.rb
veronicagg: @sarangan12 sorry, I'm not sure I understand your first comment here (in reply to Vishrut), can you clarify?
Also, what type of items are your referring to that need discussion for renaming?

sarangan12: @veronicagg Added detailed context

vishrutshah: @sarangan12 Here is my view, If you think naming sounds perfect today and we are good with that I'd keep the names as-is and not even open backlog items. If you have even a slight concern on naming, I'd like that to be fix with this changes rather than a backlog item. 

That's what I think but i'll leave it up to you.

PR #668, opened 2017-02-03
Preparing to release runtime ms_rest 0.6.3 & ms_rest_azure 0.7.0

vishrutshah:
@veronicagg Could you please review when you get a chance. Thanks! 

Also, let me know if i missed something from our yesterday's discussion. :)

### DIFF
--- a/runtime/ms_rest_azure/lib/ms_rest_azure/azure_service_client.rb
+++ b/runtime/ms_rest_azure/lib/ms_rest_azure/azure_service_client.rb
@@ -42,7 +42,7 @@ module MsRestAzure
         task = Concurrent::TimerTask.new do
           begin
             if !polling_state.azure_async_operation_header_link.nil?
-              update_state_from_azure_async_operation_header(polling_state.get_request(headers: request.headers, base_uri: request.base_uri, user_agent_extended: user_agent_extended), polling_state, custom_deserialization_block)
+              update_state_from_azure_async_operation_header(polling_state.get_request(headers: request.headers, base_uri: request.base_uri, user_agent_extended: user_agent_extended), polling_state)
             elsif !polling_state.location_header_link.nil?
               update_state_from_location_header(polling_state.get_request(headers: request.headers, base_uri: request.base_uri, user_agent_extended: user_agent_extended), polling_state, custom_deserialization_block)
             elsif http_method === :put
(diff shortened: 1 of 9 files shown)


vishrutshah:
@veronicagg Sorry but i had to update this PR. We had a bug in the recent custom_deserialization logic that was discovered as part of running existing VCR tests against regenerated latest SDKs. 

Could you please re-look into the PR when you get a chance. Thanks! 

P.S. I just realized that I could have just made another PR but ooppss!!!

vishrutshah:
@veronicagg To see what was the problem, here is the travis job

https://travis-ci.org/Azure/azure-sdk-for-ruby/jobs/199045822#L1149

runtime/ms_rest_azure/lib/ms_rest_azure/azure_service_client.rb
```diff
@@ -73,9 +73,20 @@ def get_long_running_operation_result(azure_response, custom_deserialization_blo
...
-        update_state_from_get_resource_operation(get_request, polling_state, custom_deserialization_block)
+      if AsyncOperationStatus.is_successful_status(polling_state.status)
+        # Process long-running PUT/PATCH
+        if (http_method === :put || http_method === :patch) && polling_state.resource.nil?
+          get_request = MsRest::HttpOperationRequest.new(request.base_uri, request.build_path.to_s, :get, {query_params: request.query_params, headers: request.headers})
+          update_state_from_get_resource_operation(get_request, polling_state, custom_deserialization_block)
+        end
+
+        # Process long-running POST/DELETE operation with schema defined on success status codes
+        if (http_method === :post || http_method === :delete) && custom_deserialization_block && polling_state.response
```
veronicagg: @vishrutshah could you give me a little more context on why post/delete are treated differently? what was happening for these before? what was the code path for these operations that ended up on the travis job failure?

vishrutshah: Sure, The issue was we were not handling long-running POST & DELETE operation when swagger has defined a schema on the response codes. That results in the SDK method returning nil values. To fix that, sarangan made a fixes in runtime, that caused breaking all other long-running operations [Reference: https://travis-ci.org/Azure/azure-sdk-for-ruby/jobs/199045822#L1149]. 

Now To fix that I've reverted his changes and made a simple change on how we handle POST/DELETE long-running operation. The one thing to note is that, we do not want to call get operation because these are POST/DELETE operation and that's why I wrapped them differently from PUT/PATCH (if condition just above this code).

PR #684, opened 2017-02-10
[Do Not Merge] Run bundle install --jobs=3 --retry=3 only if LATEST_RUNTIME != true

vishrutshah:


vishrutshah:
@veronicagg Sure, It's not complete yet, I was just trying out few things with travis. 

**Problem**: Travis build for `allow_failure` will fail then we are about to release azure_mgmt_* sdks and we have not released ms_rest/ms_rest_azure version as travis itself tries to run `bundle install --jobs=3 --retry=3` when `BUNDLE_GEMFILE` environment variable is set before any scripts. 

**Approach**
In this case, we should stop travis to run that command in `allow_failure` builds. 

Still working on it...

sarangan12:
OK

sarangan12:
OK

sarangan12:
OK

sarangan12:
OK

PR #807, opened 2017-06-20
Modify Rakefile to accomaodate markdown changes

sarangan12:


Rakefile
```diff
@@ -59,8 +59,12 @@ namespace :arm do
         md = REGEN_METADATA[dir.to_sym]
         ar_base_command = "#{OS.windows? ? '' : 'mono '} #{REGEN_METADATA[:autorest_loc]}"
 
-        command = "#{ar_base_command} -i #{md[:spec_uri]} -pv #{md[:version]} -n #{md[:ns]} -pn #{md[:pn].nil? ? dir : md[:pn]} -g Azure.Ruby -o lib"
-        command += " -m #{md[:modeler]}" unless md[:modeler].nil?
+        if md[:spec_uri].end_with?('.json')
+          command = "#{ar_base_command} -i #{md[:spec_uri]} -pv #{md[:version]} -n #{md[:ns]} -pn #{md[:pn].nil? ? dir : md[:pn]} -g Azure.Ruby -o lib"
+          command += " -m #{md[:modeler]}" unless md[:modeler].nil?
+        else
+          command = "#{ar_base_command} #{md[:spec_uri]} --package-version=#{md[:version]} --namespace=#{md[:ns]} --package-name=#{md[:pn].nil? ? dir : md[:pn]} --output-folder=#{File.join(Dir.pwd, 'lib')} --ruby --azure-arm"
```
veronicagg: question: was " --output-folder='lib'" not working therefore  "--output-folder=#{File.join(Dir.pwd, 'lib')}" ?

Rakefile
```diff
@@ -59,8 +59,12 @@ namespace :arm do
         md = REGEN_METADATA[dir.to_sym]
         ar_base_command = "#{OS.windows? ? '' : 'mono '} #{REGEN_METADATA[:autorest_loc]}"
 
-        command = "#{ar_base_command} -i #{md[:spec_uri]} -pv #{md[:version]} -n #{md[:ns]} -pn #{md[:pn].nil? ? dir : md[:pn]} -g Azure.Ruby -o lib"
-        command += " -m #{md[:modeler]}" unless md[:modeler].nil?
+        if md[:spec_uri].end_with?('.json')
```
veronicagg: this is fine for now, though I heard Garrett said that they are removing the legacy command line soon, when that happens, this won't work, please check with him when that will occur, so this can be updated.

Rakefile
```diff
@@ -309,10 +313,9 @@ REGEN_METADATA = {
         tag: 'arm_recovery'
     },
     azure_mgmt_recovery_services_backup: {
-        spec_uri: 'https://raw.githubusercontent.com/Azure/azure-rest-api-specs/master/arm-recoveryservicesbackup/compositeRecoveryServicesBackupClient.json',
+        spec_uri: 'https://raw.githubusercontent.com/Azure/azure-rest-api-specs/master/arm-recoveryservicesbackup/readme.md',
```
veronicagg: I assume this is the only one that has a readme.md right now, so should be fine to just update this one.

vishrutshah: It looks like you are solving this problem only for local regen story. What about the AutoPRs?

Apart from this change, we should now use the [swagger_to_sdk](https://github.com/lmazuel/swagger-to-sdk) version 2.0. I believe that it supports autorest with .json / .md.

We should refer https://github.com/Azure/azure-sdk-for-python/blob/master/swagger_to_sdk_config.json#L3 to fix that as well along with that to be consistent.

Rakefile
vishrutshah: Could you please remind me what does `--azure-arm` do?

vishrutshah: In both cases, i'd keep a value of `-o` and `--output-folder` consistent.

Rakefile
```diff
@@ -309,10 +313,9 @@ REGEN_METADATA = {
...
         ns: 'Azure::ARM::RecoveryServicesBackup',
         version: version,
-        modeler: "CompositeSwagger",
```
vishrutshah: To complete the story, we should fix  our `swagger_to_sdk_config.json` as well.

Rakefile
sarangan12: Already confirmed. Once the flip has happened and everything is fine, it will be removed eventually. So, no issues as of now.

Rakefile
sarangan12: Yes. They have changed it to relative path. So, when you have a URI as https://......./compute/xyz.md and output-folder as lib, it will try to write to https://....../compute/lib location :-(

Rakefile
sarangan12: Yes

Rakefile
sarangan12: --ruby indicates this is a ruby generator. --azure-arm indicates it is Azure Ruby generator.

Rakefile
sarangan12: That is in my To Do list. Will raise a seperate PR for that

Rakefile
sarangan12: Agreed

sarangan12:
"In both cases, i'd keep a value of -o and --output-folder consistent." -o will be removed soon in another 2 weeks.

Rakefile
vishrutshah: cool.

PR #828, opened 2017-07-05
Code changes to fix bug related to provisioning status in Polling State Object

sarangan12:
Our logic to determine Status in the Polling state object is incorrect. The [Node SDK](https://github.com/Azure/azure-sdk-for-node/blob/0a52678ea7b3a24a478975f7169fc30e9fc9759e/runtime/ms-rest-azure/lib/pollingState.js#L44) and [.NET SDK](https://github.com/Azure/azure-sdk-for-net/blob/psSdkJson6/src/SdkCommon/ClientRuntime.Azure/ClientRuntime.Azure/PollingState.cs#L49) are having similar and correct code. 

This PR is to modify our logic to ensure that it aligns with the logic of .NET and Node SDKs. 

Added rspec tests to ms_rest_azure and confirmed they are passing. Also, pointed ms_rest and ms_rest_azure locally and ran the tests in the entire sdk ruby and confirmed that nothing is breaking.

Please review and approve.

Related to https://github.com/Azure/azure-sdk-for-ruby/issues/817

### DIFF
--- a/runtime/ms_rest_azure/lib/ms_rest_azure/azure_service_client.rb
+++ b/runtime/ms_rest_azure/lib/ms_rest_azure/azure_service_client.rb
@@ -41,7 +41,7 @@ module MsRestAzure
       polling_state = PollingState.new(azure_response, @long_running_operation_retry_timeout)
       request = azure_response.request
 
-      if !AsyncOperationStatus.is_terminal_status(polling_state.status)
+      while !AsyncOperationStatus.is_terminal_status(polling_state.status)
         task = Concurrent::TimerTask.new do
           begin
             if !polling_state.azure_async_operation_header_link.nil?
--- a/runtime/ms_rest_azure/lib/ms_rest_azure/azure_service_client.rb
+++ b/runtime/ms_rest_azure/lib/ms_rest_azure/azure_service_client.rb
@@ -41,7 +41,7 @@ module MsRestAzure
       polling_state = PollingState.new(azure_response, @long_running_operation_retry_timeout)
       request = azure_response.request
 
-      while !AsyncOperationStatus.is_terminal_status(polling_state.status)
+      if !AsyncOperationStatus.is_terminal_status(polling_state.status)
         task = Concurrent::TimerTask.new do
           begin
             if !polling_state.azure_async_operation_header_link.nil?
--- a/runtime/ms_rest_azure/lib/ms_rest_azure/polling_state.rb
+++ b/runtime/ms_rest_azure/lib/ms_rest_azure/polling_state.rb
@@ -35,21 +35,35 @@ module MsRestAzure
       update_response(azure_response.response)
       @resource = azure_response.body
 
+      case @response.status
+        when 202
+          @status = AsyncOperationStatus::IN_PROGRESS_STATUS
+        when 200
+          provisioning_state = get_provisioning_state
+          @status = provisioning_state.nil?? (AsyncOperationStatus::SUCCESS_STATUS):provisioning_state
+        when 201
+          provisioning_state = get_provisioning_state
+          @status = provisioning_state.nil?? (AsyncOperationStatus::IN_PROGRESS_STATUS):provisioning_state
+        when 204
+          @status = AsyncOperationStatus::SUCCESS_STATUS
+        else
+          @status = AsyncOperationStatus::FAILED_STATUS
+      end
+    end
+
+    #
+    # Returns the provisioning status of the resource
+    #
+    # @return [String] provisioning status of the resource
+    def get_provisioning_state
       # On non flattened resource, we should find provisioning_state inside 'properties'
       if (!@resource.nil? && @resource.respond_to?(:properties) && @resource.properties.respond_to?(:provisioning_state) && !@resource.properties.provisioning_state.nil?)
-        @status = @resource.properties.provisioning_state
-      # On flattened resource, we should find provisioning_state at the top level
+        @resource.properties.provisioning_state
+        # On flattened resource, we should find provisioning_state at the top level
       elsif !@resource.nil? && @resource.respond_to?(:provisioning_state) && !@resource.provisioning_state.nil?
-        @status = @resource.provisioning_state
+        @resource.provisioning_state
       else
-        case @response.status
-          when 202
-            @status = AsyncOperationStatus::IN_PROGRESS_STATUS
-          when 200, 201, 204
-            @status = AsyncOperationStatus::SUCCESS_STATUS
-          else
-            @status = AsyncOperationStatus::FAILED_STATUS
-          end
+        nil
       end
     end
 
--- a/runtime/ms_rest_azure/lib/ms_rest_azure/polling_state.rb
+++ b/runtime/ms_rest_azure/lib/ms_rest_azure/polling_state.rb
@@ -36,14 +36,14 @@ module MsRestAzure
       @resource = azure_response.body
 
       case @response.status
-        when 202
-          @status = AsyncOperationStatus::IN_PROGRESS_STATUS
         when 200
           provisioning_state = get_provisioning_state
           @status = provisioning_state.nil?? (AsyncOperationStatus::SUCCESS_STATUS):provisioning_state
         when 201
           provisioning_state = get_provisioning_state
           @status = provisioning_state.nil?? (AsyncOperationStatus::IN_PROGRESS_STATUS):provisioning_state
+        when 202
+          @status = AsyncOperationStatus::IN_PROGRESS_STATUS
         when 204
           @status = AsyncOperationStatus::SUCCESS_STATUS
         else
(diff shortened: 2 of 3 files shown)


runtime/ms_rest_azure/lib/ms_rest_azure/azure_service_client.rb
```diff
@@ -41,7 +41,7 @@ def get_long_running_operation_result(azure_response, custom_deserialization_blo
       polling_state = PollingState.new(azure_response, @long_running_operation_retry_timeout)
       request = azure_response.request
 
-      if !AsyncOperationStatus.is_terminal_status(polling_state.status)
+      while !AsyncOperationStatus.is_terminal_status(polling_state.status)
```
sarangan12: Refer [.NET SDK Code](https://github.com/Azure/azure-sdk-for-net/blob/psSdkJson6/src/SdkCommon/ClientRuntime.Azure/ClientRuntime.Azure/AzureClientExtensions.cs#L107)  and [Node SDK Code](https://github.com/Azure/azure-sdk-for-node/blob/0a52678ea7b3a24a478975f7169fc30e9fc9759e/runtime/ms-rest-azure/lib/azureServiceClient.js#L119) for the reason of this change.

runtime/ms_rest_azure/spec/polling_state_spec.rb
```diff
@@ -92,6 +98,35 @@ module MsRestAzure
...
+          { 'Azure-AsyncOperation' => 'async_operation_header',
+            'Location' => 'location_header'})
+      azure_response = double('response',
+                              :request => nil,
+                              :response => response,
+                              :body => nil)
+
+      polling_state = PollingState.new azure_response, 0
+
+      expect(polling_state.status).to eq(AsyncOperationStatus::IN_PROGRESS_STATUS)
```
sarangan12: Earlier this used to return success status

runtime/ms_rest_azure/lib/ms_rest_azure/polling_state.rb
```diff
@@ -35,21 +35,35 @@ def initialize(azure_response, retry_timeout)
...
       if (!@resource.nil? && @resource.respond_to?(:properties) && @resource.properties.respond_to?(:provisioning_state) && !@resource.properties.provisioning_state.nil?)
-        @status = @resource.properties.provisioning_state
-      # On flattened resource, we should find provisioning_state at the top level
+        @resource.properties.provisioning_state
+        # On flattened resource, we should find provisioning_state at the top level
       elsif !@resource.nil? && @resource.respond_to?(:provisioning_state) && !@resource.provisioning_state.nil?
-        @status = @resource.provisioning_state
+        @resource.provisioning_state
       else
-        case @response.status
```
vishrutshah: I am bit confused on "what is the reason for moving this conditions into initializer rather than updating them here?"

runtime/ms_rest_azure/lib/ms_rest_azure/azure_service_client.rb
vishrutshah: Could you please elaborate on why we need while. If i understand correctly, then `Concurrent::TimerTask` should do polling unitl success on interval of `polling_delay`

runtime/ms_rest_azure/lib/ms_rest_azure/polling_state.rb
```diff
@@ -35,21 +35,35 @@ def initialize(azure_response, retry_timeout)
       update_response(azure_response.response)
       @resource = azure_response.body
 
+      case @response.status
```
vishrutshah: I'd not rely on that the recorded test passed because I know that we are modifying the polling in the recording. It'd be great if you can remove recordings or some VM creation and Network / the customer scenario to make sure we are good.

runtime/ms_rest_azure/lib/ms_rest_azure/polling_state.rb
```diff
@@ -35,21 +35,35 @@ def initialize(azure_response, retry_timeout)
...
+        when 202
```
veronicagg: minor: could we reorder cases to have 200 and 201, then 202 and 204?

runtime/ms_rest_azure/spec/polling_state_spec.rb
```diff
@@ -10,25 +10,31 @@ module MsRestAzure
   describe PollingState do
     it 'should initialize status from flattened response body' do
       response_body = double('response_body', :provisioning_state => 'InProgress')
-      response = double('response',
```
veronicagg: was the test before these changes valid? By this question I mean, would it be valid to have a nil "response" with a response body?

veronicagg: could there be some type of inconsistency between the response content and provisioning state, and if there is, how does our code handle it? we should be returning some kind of error if something does not match up, but try to avoid failing with a null exception or similar, so just something to check on.

runtime/ms_rest_azure/spec/polling_state_spec.rb
```diff
@@ -10,25 +10,31 @@ module MsRestAzure
...
+      polling_state = PollingState.new azure_response, 0
 
       expect(polling_state.status).to eq('InProgress')
     end
 
     it 'should initialize status from non-flattened response body' do
       provisioning_state = double('provisioning_state', :provisioning_state => 'Succeeded')
       response_body = double('response_body', :properties => provisioning_state)
-      response = double('response',
+      response = double('response', :status => 201, :headers =>
```
veronicagg: should this be 201 or 200 instead?

runtime/ms_rest_azure/spec/polling_state_spec.rb
veronicagg: based on the guidelines here https://github.com/Azure/azure-resource-manager-rpc/blob/master/v1.0/Addendum.md#provisioningstate-property , no provisioning state means success, so I'm not sure making it "in progress" is what's expected. I'm not sure if it's a valid response to return 201 and no provisioning state either (so maybe we never get into this case?), could it be that the only accepted code for not sending provisioning state is 200?

runtime/ms_rest_azure/lib/ms_rest_azure/polling_state.rb
vishrutshah: Yup you both are right. I did misread it :)

runtime/ms_rest_azure/lib/ms_rest_azure/polling_state.rb
```diff
@@ -35,21 +35,35 @@ def initialize(azure_response, retry_timeout)
...
       else
-        case @response.status
-          when 202
-            @status = AsyncOperationStatus::IN_PROGRESS_STATUS
-          when 200, 201, 204
-            @status = AsyncOperationStatus::SUCCESS_STATUS
-          else
-            @status = AsyncOperationStatus::FAILED_STATUS
-          end
+        nil
```
sarangan12: **is returning nil in the "else" case maintaining previous behavior?**
Yes. It is maintaining the previous behavior

runtime/ms_rest_azure/spec/polling_state_spec.rb
sarangan12: **was the test before these changes valid? By this question I mean, would it be valid to have a nil "response" with a response body?**
No. A nil response is not a valid response. Think of the scenario that you call a web service. It should return something with some status. Else, it is a timeout

runtime/ms_rest_azure/spec/polling_state_spec.rb
sarangan12: It is 201.

runtime/ms_rest_azure/lib/ms_rest_azure/polling_state.rb
sarangan12: Done

runtime/ms_rest_azure/lib/ms_rest_azure/polling_state.rb
sarangan12: No action here

runtime/ms_rest_azure/spec/polling_state_spec.rb
sarangan12: Discussed with @devigned These services have created before the arm guidelines. That was the reason .NET SDK and Node SDK follow the same logic and with this PR we are aligning our code with them

runtime/ms_rest_azure/lib/ms_rest_azure/polling_state.rb
sarangan12: Agreed. Ran the cases with deleting a file from network and compute and confirmed it is successfully tested.

runtime/ms_rest_azure/spec/polling_state_spec.rb
veronicagg: sounds reasonable.

runtime/ms_rest_azure/lib/ms_rest_azure/polling_state.rb
veronicagg: do we need to re-record any tests?

runtime/ms_rest_azure/lib/ms_rest_azure/polling_state.rb
devigned: If we don't understand a case, it would probably be best to just keep polling. I believe status is unbounded, so we should not expect to understand all of them.

Issue #610, opened 2016-12-01
Refactor Serialization


devigned:
The client really doesn't need to be involved in serialization. The client coupled to serialization only because the deserializer needs to know how to retrieve the model type with relationship to the root of the client's module. If the serialization logic is decorated on the model, this becomes trivial; insatiate a `self.class`.

```ruby
      # Serialize Request
      request_mapper = VirtualMachine.mapper()
      request_content = @client.serialize(request_mapper,  parameters, 'parameters')
      request_content = request_content != nil ? JSON.generate(request_content, quirks_mode: true) : nil
      # ....
      result_mapper = VirtualMachine.mapper()
      result.body = @client.deserialize(result_mapper, parsed_response, 'result.body')
```
To:
```ruby
      # Serialize Request
      request_content = VirtualMachine.serialize(parameters, name: 'parameters', quirks_mode: true)
      # ....
      result.body = VirtualMachine.deserialize(parsed_response, name: 'result.body')
```

Discussed here: https://github.com/Azure/azure-sdk-for-ruby/pull/599

PR #860, opened 2017-07-25
Code changes for serialization/deserialization

sarangan12:
This is a minor bug fix. Currently no code uses this. Please review and approve

### DIFF
--- a/runtime/ms_rest/lib/ms_rest/jsonable.rb
+++ b/runtime/ms_rest/lib/ms_rest/jsonable.rb
@@ -13,8 +13,9 @@ module MsRest
     # @return [String] JSON serialized version of the object
     #
     def to_json(options = nil)
-      mapper = self.class.mapper if options.nil? || !options.key?(:mapper)
-      serialize(mapper, self)
+      mapper = (options.nil? || !options.key?(:mapper))? self.class.mapper: options[:mapper]
+      object = (options.nil? || !options.key?(:object))? self: options[:object]
+      serialize(mapper, object)
     end
 
     #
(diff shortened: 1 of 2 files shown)


runtime/ms_rest/lib/ms_rest/jsonable.rb
```diff
@@ -13,8 +13,9 @@ module JSONable
     # @return [String] JSON serialized version of the object
     #
     def to_json(options = nil)
-      mapper = self.class.mapper if options.nil? || !options.key?(:mapper)
-      serialize(mapper, self)
+      mapper = (options.nil? || !options.key?(:mapper))? self.class.mapper: options[:mapper]
+      object = (options.nil? || !options.key?(:object))? self: options[:object]
```
veronicagg: what would be passed in the ":object" option? What's the reason for introducing this? 
Would the result be proper if there's no ":mapper" option, but there's ":object" option passed?

sarangan12:
@veronicagg @vishrutshah This is done to handle the issue https://github.com/Azure/azure-sdk-for-ruby/issues/610. 

In order to use the JSONable, we need to be able to pass both the object and mapper. Both are optional. You can pass one or both and the default values will be used. 

How will the resulting code look:

For composite models, it will be like:
```
Azure::ARM::Compute::Models::AvailabilitySet.to_json #Here you dont have to pass the mapper/object. 
```
For non composite models, it will be like:
```
def put_utc_max_date_time_async(datetime_body, custom_headers = nil)
..........
..........
request_mapper = {
        required: true,
        serialized_name: 'datetimeBody',
        type: {
          name: 'DateTime'
        }
      }
current_module = Object.const_get(self.class.to_s.split( '::' )[0, self.class.to_s.split( '::' ).length-1].join)
a_new_class = Class.new { include MsRest::JSONable }
current_module.const_set('ANewClass', a_new_class)
request_content = a_new_class.new.to_json({mapper: request_mapper, object: datetime_body })
..........
..........
```

veronicagg:
@sarangan12 ok, should we be adding a test (or two) in ms_rest to check on this added functionality? It can help not only validate but describe the usage.

sarangan12:
@veronicagg Why stop with 2? :-) added 3 cases 1 with mapper (without object), 1 with object (without mapper), 1 with object and mapper

Issue #900, opened 2017-08-17
Name parameter is required (Data Lake Store Account)


muhammad-asad-confiz:
Hi,
I am using `azure_mgmt_datalake_store`. I want to report that in `Firewall Rule` object, the name parameter is marked as an optional parameter [here](https://github.com/Azure/azure-sdk-for-ruby/blob/75f0b3f93e7fe779b0e7082db7d89b75b7d6d368/management/azure_mgmt_datalake_store/lib/generated/azure_mgmt_datalake_store/models/firewall_rule.rb#L45) but it is required.
Please make it required in the documentation.

veronicagg:
@muhammad-asad-confiz I believe the name parameter in this case comes from https://github.com/Azure/azure-rest-api-specs/blob/current/specification/datalake-store/resource-manager/Microsoft.DataLakeStore/2016-11-01/account.json#L1573
Feel free to open an issue in the https://github.com/Azure/azure-rest-api-specs repo to evaluate updates to the specification for this property. If the service team approved the update, it can then be merged and would transfer to the generated SDKs.

veronicagg:
I'll open an issue in the azure-rest-api-spec repo for service team evaluation

veronicagg:
Please track this issue in the referenced issue, as we would need to pick up this change from the specification to reflect it in the generated sdk.

PR #923, opened 2017-08-24
Name parameter is required (Data Lake Store Account)

muhammad-asad-confiz:
Hi,
I am using azure_mgmt_datalake_store. I want to report that in Firewall Rule object, the name parameter is marked as an optional parameter here but it is required. I have made the changes. #900


veronicagg:
@muhammad-asad-confiz I believe the name parameter in this case comes from https://github.com/Azure/azure-rest-api-specs/blob/current/specification/datalake-store/resource-manager/Microsoft.DataLakeStore/2016-11-01/account.json#L1573 
Feel free to open an issue in the https://github.com/Azure/azure-rest-api-specs repo to evaluate updates to the specification for this property. If the service team approved the update, it can then be merged and would transfer to the generated SDKs.

Issue #881, opened 2017-08-09
Multi-api-version support: Update unit test to match code changes


veronicagg:
Update unit tests to match code changes and validate latest versions of the specs.

veronicagg:
Make sure to add appropriate folder under "spec" folder to account for the corresponding api-version.

veronicagg:
https://github.com/Azure/azure-sdk-for-ruby/pull/925

Issue #880, opened 2017-08-09
Multi-api-version support: Update directory structure


veronicagg:
Update directory structure for code generation to account for multiple api versions for each service.

Issue #879, opened 2017-08-09
Multi-api-version support: Update code generation to include api-version in namespaces


veronicagg:
Need to update code generation to take into account API versions in namespaces, and use the fully qualified class names for model creation in generated code.

PR #925, opened 2017-08-25
Azure stack multi api

veronicagg:
Creating a PR against azure-stack branch to show:
1. Sample SDK generation (with my updated autorest code) which includes api versions
2. Updated folder structure
3. Metadata file for autorest arguments (config.json)

Please let me know your feedback.

(related to https://github.com/Azure/azure-sdk-for-ruby/issues/880, https://github.com/Azure/azure-sdk-for-ruby/issues/879, https://github.com/Azure/azure-sdk-for-ruby/issues/920, https://github.com/Azure/azure-sdk-for-ruby/issues/881)


vishrutshah:
Overall structure looks awesome as we discussed 💯 

Here are some general questions:

- [ ] Top level file `azure_mgmt_xxx.rb` in `lib` folder - Which api-version's sub-level `xxx-xxx-xxx\azure_mgmt_xxx.rb` file should be auto loaded? 
- [ ] Where should `version.rb` file live for the bundle/gem of `azure_mgmt_xxxx` ?
- [ ] Should [module_definition.rb](https://github.com/veronicagg/azure-sdk-for-ruby/blob/azure-stack-multi-api/management/azure_mgmt_compute/lib/2017-03-30/generated/azure_mgmt_compute/module_definition.rb#L8) change with new level of api-version? 
- [ ] Something seems off in [azure_mgmt_resources](https://github.com/veronicagg/azure-sdk-for-ruby/tree/azure-stack-multi-api/management/azure_mgmt_resources/lib) folder
- [ ] Would it be useful to send the api-version in telemetry [here ](https://github.com/veronicagg/azure-sdk-for-ruby/blob/azure-stack-multi-api/management/azure_mgmt_compute/lib/2017-03-30/generated/azure_mgmt_compute/module_definition.rb#L8)?
- [ ] How does the intelliSense look? I am curious to know as I have not idea whether we need something on that line or not :)
 
Some minor improvements into AutoRest template would be needed:
- [ ] Fixing Indentation
       Class : [AvailabilitySets](https://github.com/veronicagg/azure-sdk-for-ruby/blob/azure-stack-multi-api/management/azure_mgmt_compute/lib/2016-03-30/generated/azure_mgmt_compute/availability_sets.rb#L11)
        AutoLoad: [Here](https://github.com/veronicagg/azure-sdk-for-ruby/blob/azure-stack-multi-api/management/azure_mgmt_compute/lib/2016-03-30/generated/azure_mgmt_compute.rb#L24)
- [ ] Find place for `version.rb` file

Let me know your thoughts ! Thanks!

veronicagg:
@vishrutshah Thanks for the feedback and questions:
1. This file may need to load the generated client we're going to produce which allows putting together all versions and default to one of them. 
2. We may need to discuss this one, as I understand it the version should be for the full gem, so we could have this one at the "new generated client" level.
3. I wasn't sure about this, so I left it untouched, what's the benefit of updating this file?
4. Sorry, could you expand? azure_mgmt_resource is not listed in the config.json, so nothing got generated for it at the moment.
5. Didn't think of it, but probably yes.
6. Once the proper module is included, intellisense looks the same as before. If including the model, doesn't specify the api-version, intellisene shows the options from each api-version. I think the intellisense aspect comes more in place once the uber client is in place and when using the services via Profiles. 

Re: Indentation - I didn't put much focus on that, yes, I'll need to review, thanks.
Re: version.rb - see above.

Let me know what you think, thanks!

veronicagg:
Made multiple commits to hopefully make it easier to review. 

- I've added version.rb to each api-version, though most likely this should be a top level file. For now, I pointed the gemspec version to the constant in the specific api-version'd file.
- I've updated module_definition file.
- I made updates to the referenced paths in the generated/azure_mgmt_*.rb files
- path in non-generated file lib/azure_mgmt_*rb also updated to one api-version, I believe this would need to be updated once we have a client at the top level.
- Have not updated telemetry yet
- RE: Indentation - code now includes the namespace update at the top, coming from namespace passed into autorest as a parameter, so not adding "module" lines in the files. I did have to modify autorest to the get paths in the above mentioned files updated. Will send a PR to autorest, once I pull the latest changes from there and port my changes from the "old" autorest style to the new generators-split code.
- I've also updated one of the resource management tests to show the type of updates needed, and the test executes. I haven't yet put the tests into a specific api-version folder.

Rakefile
```diff
@@ -50,20 +52,36 @@ namespace :arm do
...
+  desc 'Regen code for each sdk with all its api versions'
+  task :apiVersions => :clean_generated_api_versions do
+    json = get_config_file
     REGEN_EXCLUDES = ['azure_sdk']
-    each_gem do |dir|
+    each_gem do |dir| # dir corresponds to each azure_mgmt_* folder
       unless REGEN_EXCLUDES.include?(dir.to_s)
         puts "\nGenerating #{dir}\n"
-        md = REGEN_METADATA[dir.to_sym]
         ar_base_command = "#{OS.windows? ? '' : 'mono '} #{REGEN_METADATA[:autorest_loc]}"
```
vishrutshah: minor: With new autorest with npm one, we may not need to have this conditions `OS.windows? ? '' : 'mono '`

Rakefile
```diff
@@ -108,6 +126,11 @@ def execute(cmd)
   end
 end
 
+def get_config_file
+  config_file = File.read "F:/Git/azure-sdk-for-ruby/config.json"
```
vishrutshah: minor: let's put relative path here

Rakefile
```diff
@@ -50,20 +52,36 @@ namespace :arm do
...
+             if argument_name.casecmp("markdown") == 0
+               ar_arguments = ar_arguments + " #{argument_value}"
+             else
+               ar_arguments = ar_arguments + " --#{argument_name}=#{argument_value}"
+             end
+           end
+         end
+         # puts "Arguments: #{ar_arguments}"
+         command = "#{ar_base_command} --package-name=#{package_name} #{ar_arguments} --output-folder=#{File.join(Dir.pwd, 'lib', output_folder)} --ruby --azure-arm"
+         puts "command: #{command}"
```
vishrutshah: minor: we can remove this line as `execute_and_stream` would print the command it is running..

Rakefile
```diff
@@ -122,229 +145,18 @@ end
 
 def each_gem
   each_child do |dir|
-    gem_dir = dir.split('/').last.to_sym
-    if REGEN_METADATA.has_key?(gem_dir)
+    #puts "Each child #{dir}"
+    gem_dir = dir.split('/').last
```
vishrutshah: curious: why not use `.to_sym` while looking up in the dictionary ?

Rakefile
```diff
@@ -122,229 +145,18 @@ end
...
+    if get_config_file.has_key?(gem_dir)
+     # puts "Gem dir #{gem_dir}"
       yield gem_dir
     end
   end
 end
 
 REGEN_METADATA = {
-    autorest_loc: ENV.fetch('AUTOREST_LOC', '../../../autorest/binaries/net45/AutoRest.exe'),
+    autorest_loc: "node F:/Git/newautorest/autorest/src/autorest-core/dist/app.js", #autorest", #ENV.fetch('AUTOREST_LOC', '../../../autorest/binaries/net45/AutoRest.exe'),
```
vishrutshah: let's look up the environment variable or assume that `autorest` is in the path

Rakefile
```diff
@@ -122,229 +145,18 @@ end
...
     azure_sdk: {
```
vishrutshah: Should we move this to the config file as well and get rid off the REGEN_METADATA?

config.json
```diff
@@ -2,87 +2,85 @@
     "azure_mgmt_analysis_services": {
         "azure_mgmt_analysis_services_2016_05_16":{
             "input-file": "https://raw.githubusercontent.com/Azure/azure-rest-api-specs/current/specification/analysisservices/resource-manager/Microsoft.AnalysisServices/2016-05-16/analysisservices.json",
-            "namespace": "Azure::ARM::AnalysisServices",
-            "version": "0.11.0",
-            "output-folder": "2016-05-16"
+            "namespace": "Azure::ARM::AnalysisServices::Api_2016_05_16",
+            "package-version": "0.11.0"
```
vishrutshah: We may be able to generalize `package-version` to one level up in the dictionary so that we only override for the different versions if you like

config.json
```diff
@@ -2,87 +2,85 @@
...
+            "namespace": "Azure::ARM::Compute::Api_2017_03_30",
+            "package-version": "0.11.0",
             "tag": "package-compute-2017-03"
         },
         "azure_mgmt_compute_2016_04_preview": {
             "markdown": "https://raw.githubusercontent.com/Azure/azure-rest-api-specs/current/specification/compute/resource-manager/readme.md",
-            "namespace": "Azure::ARM::Compute",
-            "version": "0.11.0",
-            "output-folder": "2016-04-preview",
+            "namespace": "Azure::ARM::Compute::Api_2016_04_preview",
```
vishrutshah: Curious: Here we are defining `Azure::ARM::Compute::Api_xxxx_xx_preview` that is different from api version value `2016-04-30-preview` is that expected? That seems folder name in generated code is different from namespace. I am bit confused here whether this is expected or not.

management/azure_mgmt_compute/azure_mgmt_compute.gemspec
```diff
@@ -5,12 +5,12 @@
 lib = File.expand_path('../lib', __FILE__)
 $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
 
-require 'generated/azure_mgmt_compute/module_definition'
-require 'generated/azure_mgmt_compute/version'
+require '2017-03-30/generated/azure_mgmt_compute/module_definition'
```
vishrutshah: This is pretty cool. How did you do that is selects the latest version here and in the version as well :)

management/azure_mgmt_compute/lib/2016-03-30/generated/azure_mgmt_compute.rb
```diff
@@ -16,133 +16,131 @@
...
-  autoload :VirtualMachineExtensions,                           'generated/azure_mgmt_compute/virtual_machine_extensions.rb'
-  autoload :VirtualMachineImages,                               'generated/azure_mgmt_compute/virtual_machine_images.rb'
-  autoload :UsageOperations,                                    'generated/azure_mgmt_compute/usage_operations.rb'
-  autoload :VirtualMachineSizes,                                'generated/azure_mgmt_compute/virtual_machine_sizes.rb'
-  autoload :VirtualMachines,                                    'generated/azure_mgmt_compute/virtual_machines.rb'
-  autoload :VirtualMachineScaleSets,                            'generated/azure_mgmt_compute/virtual_machine_scale_sets.rb'
-  autoload :VirtualMachineScaleSetVMs,                          'generated/azure_mgmt_compute/virtual_machine_scale_set_vms.rb'
-  autoload :ContainerServices,                                  'generated/azure_mgmt_compute/container_services.rb'
-  autoload :ComputeManagementClient,                            'generated/azure_mgmt_compute/compute_management_client.rb'
+module Azure::ARM::Compute::Api_2017_01_31
```
vishrutshah: This seems odd. Namespace supposed to be `Azure::ARM::Compute::Api_2016_03_30`. Isn't it?

management/azure_mgmt_compute/lib/azure_mgmt_compute.rb
```diff
@@ -2,4 +2,4 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License. See License.txt in the project root for license information.
 
-require 'generated/azure_mgmt_compute'
\ No newline at end of file
+require '2017-03-30/generated/azure_mgmt_compute'
```
vishrutshah: 👍 👍

Rakefile
```diff
@@ -28,10 +29,11 @@ namespace :arm do
     end
   end
 
-  desc 'Delete ./lib/generated for each of the Azure Resource Manager projects'
-  task :clean_generated do
+  desc 'Delete ./lib/generated for each sdk'
+  task :clean_generated_api_versions do
     each_gem do
-      execute_and_stream(OS.windows? ? 'del /S /Q lib\generated 2>nul' : 'rm -rf lib/generated')
+      delete_command_windows = 'for /r lib\ %a in (generated) do @if exist %a del /S /Q %a 2>nul'
```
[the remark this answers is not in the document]
veronicagg: yes, I couldn't get the previous command to work, so I looked up online and found this option that did. I will try again and see if the previously used command works now.

Rakefile
```diff
@@ -28,10 +29,11 @@ namespace :arm do
...
+      execute_and_stream(OS.windows? ? delete_command_windows : 'rm -rf lib/*/generated')
```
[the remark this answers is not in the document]
veronicagg: sure, I can add that, I haven't tried the linux command yet, so that's pending.

Rakefile
veronicagg: that would be nice, we just need to try it on the mac, I haven't yet.

Rakefile
veronicagg: yes, I can remove most of the puts :) I added them for my debugging purposes and just left them in for now.

Rakefile
veronicagg: yes, definitely! As I do my next round of updates, will get this updated.

Rakefile
veronicagg: we used to look for the symbol when we had the dictionary defined in the rakefile, now when loading the json file (config.json), these values are not loaded as symbols but strings. We can look into this further, if you think there could be benefits of having symbols.

Rakefile
veronicagg: yes, definitely! this can just be "autorest" as default, since that's how the new stuff works. I'll update it in my next round, as I'm now on the latest autorest code, and just published a PR to the autorest.ruby repo

config.json
veronicagg: We could, I left it this way to simplify things for now. If we think there's value on globalizing some values, we can add logic to look for these values. Let's discuss this as we put all things together and figure out if it's needed.

management/azure_mgmt_compute/azure_mgmt_compute.gemspec
veronicagg: Hehe, the gemspec file is not autogenerated at the moment, as it wasn't before, so I updated it. Let's evaluate whether we want to to this automatically and where, because we don't generate this via autorest today, maybe we can do it with the "uber" client generation.

management/azure_mgmt_compute/lib/2016-03-30/generated/azure_mgmt_compute.rb
veronicagg: Good catch, this is because I messed up here; https://github.com/Azure/azure-sdk-for-ruby/pull/925/files#diff-06b2d3b23dce96e1619d2b53d6c947ecR74 
this stuff comes from configuration, so if we make a mistake there, we'll see it here. We'll have to review the metadata file (manual or autogenerated)

management/azure_mgmt_compute/lib/azure_mgmt_compute.rb
veronicagg: this is another file I edited manually, I think this file needs to change as we introduce an "uber" client for the gem. We should discuss this.

veronicagg:
@sarangan12 I've pushed updated generated code, please let me know if you see more inconsistencies.

veronicagg:
@vishrutshah @sarangan12 I believe I've addressed all the feedback, please let me know if you see anything missing. I've included all the gems that have tests in the config.json and generated them, as well as updated their unit tests. 
There's a pending question for packaging of network gem, since I don't know where we should be shipping the files that are not of the same api-version - see https://github.com/Azure/azure-rest-api-specs/issues/1685 
Also, the config file should probably include more versions, I included at least the latest, I'd like to know which ones we should really be shipping - needs review.

management/azure_mgmt_compute/lib/azure_mgmt_compute.rb
vishrutshah: I forgot whether we decided to only include one api-version here or all of them.

vishrutshah:
@veronicagg Overall changes look great. Just left few question for us to decide on. Let me know if you want to chat

veronicagg:
Updated generation and added/split container services in a separate gem as I see that python is doing it this way https://github.com/Azure/azure-sdk-for-python/tree/master/azure-mgmt-containerservice

Rakefile
```diff
@@ -53,15 +57,14 @@ namespace :arm do
...
   desc 'Regen code for each sdk with all its api versions'
-  task :apiVersions => :clean_generated_api_versions do
+  task :regen => :clean_generated do
     json = get_config_file
     REGEN_EXCLUDES = ['azure_sdk']
     each_gem do |dir| # dir corresponds to each azure_mgmt_* folder
       unless REGEN_EXCLUDES.include?(dir.to_s)
         puts "\nGenerating #{dir}\n"
-        ar_base_command = "#{OS.windows? ? '' : 'mono '} #{REGEN_METADATA[:autorest_loc]}"
+        ar_base_command = "#{ENV.fetch('AUTOREST_LOC', 'autorest')}" + " --use=F:/Git/autorest.ruby"
```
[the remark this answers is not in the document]
veronicagg: I agree, sorry, I forgot and left that in there!

config.json
```diff
@@ -1,86 +1,189 @@
...
+            "namespace": "Azure::ARM::Network::Api_2017_09_01",
+            "package-version": "0.11.0"
+        },
+        "azure_mgmt_network_2017_03_30":{
+            "input-file": ["https://raw.githubusercontent.com/Azure/azure-rest-api-specs/current/specification/network/resource-manager/Microsoft.Network/2017-09-01/vmssNetworkInterface.json",
+                            "https://raw.githubusercontent.com/Azure/azure-rest-api-specs/current/specification/network/resource-manager/Microsoft.Network/2017-09-01/vmssPublicIpAddress.json"],
+            "namespace": "Azure::ARM::Network::Api_2017_03_30",
+            "package-version": "0.11.0"
+        }//,
+        // "azure_mgmt_network_2017_08_01":{
```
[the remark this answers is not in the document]
veronicagg: I'm not sure how we're going to package network (that's one of my pending questions - since there are 2 files in the packages defined in readme.md with different versions, and they actually belong to compute - see https://github.com/Azure/azure-rest-api-specs/issues/1685), if one of the sets looks correct, i'll add the rest.

config.json
```diff
@@ -1,86 +1,189 @@
...
         "azure_mgmt_resources_2017_05_10":{
-            "input-file": "https://raw.githubusercontent.com/Azure/azure-rest-api-specs/current/specification/resources/resource-manager/Microsoft.Resources/2017-05-10/resources.json",
+            "input-file": ["https://raw.githubusercontent.com/Azure/azure-rest-api-specs/current/specification/resources/resource-manager/Microsoft.Resources/2017-05-10/resources.json"],
             "namespace": "Azure::ARM::Resources::Api_2017_05_10",
             "package-version": "0.11.0"
+        },
+        "azure_mgmt_resources_2016_09_01":{
+            "input-file": ["https://raw.githubusercontent.com/Azure/azure-rest-api-specs/current/specification/resources/resource-manager/Microsoft.Resources/2016-09-01/resources.json"],
+            "namespace": "Azure::ARM::Resources::Api_2016_09_01",
+            "package-version": "0.11.0"
```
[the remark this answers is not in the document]
veronicagg: once we decide on the version I can bulk update all, at that point we'll have to write a change log as well.

management/azure_mgmt_analysis_services/azure_mgmt_analysis_services.gemspec
```diff
@@ -5,12 +5,12 @@
...
 
-require '2016-05-16/generated/azure_mgmt_analysis_services/module_definition'
-require '2016-05-16/generated/azure_mgmt_analysis_services/version'
+require 'module_definition'
+require 'version'
 
 Gem::Specification.new do |spec|
   spec.name          = 'azure_mgmt_analysis_services'
-  spec.version       = Azure::ARM::AnalysisServices::Api_2016_05_16::VERSION
+  spec.version       = Azure::ARM::AnalysisServices::VERSION
```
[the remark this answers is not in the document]
veronicagg: not sure I understand what you mean

management/azure_mgmt_compute/lib/azure_mgmt_compute.rb
veronicagg: i'll see if I need to add other api versions, via running tests for another version too.

vishrutshah: from what I see in the below article, I think we may need all of them here.

>>>When packaging your gem files, you need to be careful of what is in your lib directory. Every gem you have installed gets its lib directory appended onto your $LOAD_PATH. This means any file on the top level of the lib directory could get required

Seems like that is how we expose our gems into $LOAD_PATH to be used by `require "xxx"`

http://guides.rubygems.org/patterns/

veronicagg: yes, I think we will need them. I tried resources with tests for another api version and needed to add it, so I'll be adding it to others.

management/azure_mgmt_analysis_services/azure_mgmt_analysis_services.gemspec
```diff
@@ -5,8 +5,8 @@
 lib = File.expand_path('../lib', __FILE__)
 $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
 
-require 'generated/azure_mgmt_analysis_services/module_definition'
-require 'generated/azure_mgmt_analysis_services/version'
+require_relative 'lib/module_definition'
```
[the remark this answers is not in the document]
veronicagg: thanks for the pointer, I think I'll stick with "require" for now to be consistent, and open an issue to tackle "require_relative" and loading enhancement, mainly because 1) autorest generator would need to be re-adjusted, 2) not sure if the LOAD_PATHs we have are appropriate or need to be adjusted as well. Does this sound good?

veronicagg:
Closed and reopened to see if PR picks up bundler update for older api versions.

veronicagg:
@vishrutshah @sarangan12 I believe I addressed all the feedback. There's only one linux command to delete directories missing, that I'd like to do in a separate PR.
If this looks good, can we get it merged to the azure-stack branch?
For now I kept all the gems, pending on response from PM.

veronicagg:
updated linux command to clean generated folders, thanks @vishrutshah .

veronicagg:
I'll be updating readme files for the gems, once we have the profile per gem definition, so I can include both options in it, via profile or specifying the api-version.

veronicagg:
@vishrutshah I'll finalize config.json tomorrow, as I got confirmation today on what to ship. Particularly, we will not be shipping api_management gem at all per request. 
Regarding the comment about "generated from autorest" at the top of the file, you're right, I'll fix that, I'll do it in another PR, since, I'd prefer to merge this one, so @sarangan12 can more easily use it.

PR #972, opened 2017-09-12
[rake arm:release] Rescue error while publishing

vishrutshah:
Swallowing errors while publishing to allow publishing of the only new version of the gems

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -45,8 +45,12 @@ namespace :arm do
   desc 'Push gem for each of the Azure Resource Manager projects'
   task :release, [:key] => :build do |_, args|
     each_gem do |dir|
-      # Using execute method so that keys are not logged onto console
-      execute("gem push ./pkg/#{dir}-#{version}.gem" + (args[:key].nil? ? '' : " -k #{args[:key]}"))
+      begin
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -45,8 +45,12 @@ namespace :arm do
   desc 'Push gem for each of the Azure Resource Manager projects'
   task :release, [:key] => :build do |_, args|
     each_gem do |dir|
-      # Using execute method so that keys are not logged onto console
-      execute("gem push ./pkg/#{dir}-#{version}.gem" + (args[:key].nil? ? '' : " -k #{args[:key]}"))
+      begin
+        # Using execute method so that keys are not logged onto console
+        execute("gem push ./pkg/#{dir}-#{version}.gem" + (args[:key].nil? ? '' : " -k #{args[:key]}"))
+      rescue Exception => ex
+        puts "Error occurred while publishing #{dir}: #{ex}"
```
veronicagg: I would do caps on "ERROR"

vishrutshah: that's true but i'll revert this changes once we release :)

vishrutshah:
Thanks for the quick review!!

PR #973, opened 2017-09-12
Revert "[rake arm:release] Rescue error while publishing (#972)"

vishrutshah:
This reverts commit 59eeea598daf1868637ef979fab8bef33534d49d.

The reason we added this change was to make sure even one of the gem publish fails we can still try releasing other gems. This was a temporary change but once we have publishing story for individual gem we'll be able to better control how we release an each gem. 

cc: @sarangan12 Let's talk about this in case you were working on release story task :)


vishrutshah:
Thanks again for the quick review!!

PR #978, opened 2017-09-15
Updating bundler version for travis CI

veronicagg:
CI is failing for PR https://github.com/Azure/azure-sdk-for-ruby/pull/925 for older ruby versions, it appears that travis is running with older versions of bundler. 
Older version is running with 1.8 version and we need fix https://github.com/bundler/bundler/commit/f4481a7e3dbaf7f563e32afbe2cbbb84e3091f16 which was included starting 1.10.

Making bundler version update same as https://github.com/Azure/azure-ruby-asm-core/blob/master/.travis.yml#L8

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ rvm:
   - 2.1.5
   - 2.2.0
   - 2.3.0
-before_install: gem install bundler -v 1.14.1
 matrix:
   include:
     - rvm: 2.3.0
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -4,6 +4,7 @@ rvm:
   - 2.1.5
   - 2.2.0
   - 2.3.0
+before_install: gem install bundler -v 1.14.1
```
vishrutshah: minor: let's put this line between line number 15 and 16 to have one section of `before_install`

PR #1009, opened 2017-09-29
Changes related to fixing the inheritance logic.

sarangan12:
As discussed offline, this PR is related to https://github.com/Azure/autorest.ruby/pull/13 and fixes the tests associated with those changes. 

@vishrutshah @veronicagg @salameer Please review and approve


management/azure_mgmt_compute/spec/virtual_machines_spec.rb
```diff
@@ -113,11 +113,11 @@
     ni = network_client.make_request(:get, ni_path, options)
     expect(ni).not_to be_nil
     # user NetworkInterface mapper to deserialize object
-    ni_instance  = NetworkInterface.new.from_json(ni)
+    ni_instance  = network_client.deserialize(NetworkInterface.mapper(), ni)
     # retrieve first IP address for first IP configuration
     ip_address_path = ni_instance.ip_configurations[0].public_ipaddress.id
     ip_address = network_client.make_request(:get, ip_address_path, options)
-    ip_address_instance = PublicIPAddress.new.from_json(ip_address)
+    ip_address_instance = network_client.deserialize(PublicIPAddress.mapper(), ip_address)
```
vishrutshah: just curious what has caused this change?

management/azure_mgmt_service_bus/spec/queues_spec.rb
```diff
@@ -21,7 +21,6 @@
     namespace_parameters = Azure::ARM::ServiceBus::Models::SBNamespace.new
     namespace_parameters.location = 'westus'
     queue_parameters = Azure::ARM::ServiceBus::Models::SBQueue.new
-    queue_parameters.location = 'westus'
```
vishrutshah: any reason for removing this? Just wanted to make sure what has caused this?

management/azure_mgmt_compute/spec/virtual_machines_spec.rb
sarangan12: I was working with David on a different task. It was dropped eventually. This was a remnant from that task

management/azure_mgmt_service_bus/spec/queues_spec.rb
sarangan12: Because accroding to the Resource definition in ServiceBus queue, there is no location. https://github.com/Azure/azure-rest-api-specs/blob/current/specification/servicebus/resource-manager/Microsoft.ServiceBus/2017-04-01/servicebus.json#L2503

sarangan12:
@veronicagg The CI will pass as it is running against the current code.

sarangan12:
@vishrutshah Changed the expect per offline discussion

management/azure_mgmt_compute/spec/virtual_machines_spec.rb
```diff
@@ -135,7 +135,7 @@
 
     result = @resource_helper.network_client.network_interfaces.list_all
     result.each do |network_interface|
-      expect(network_interface.virtual_machine).to be_an_instance_of(MsRestAzure::SubResource)
+      expect(network_interface.virtual_machine).to be_an_instance_of(VirtualMachine)
```
vishrutshah: you should have to write `Azure::ARM::Compute::Models::VirtualMachine`

sarangan12: Actually, it should be Azure::ARM::Network::Models::SubResource. The model looks like this:

```
class NetworkInterface < Resource

      include MsRestAzure

      # @return [SubResource] The reference of a virtual machine.
      attr_accessor :virtual_machine
```

So, changed it with the latest commit

management/azure_mgmt_analysis_services/lib/generated/azure_mgmt_analysis_services/models/analysis_services_server_update_parameters.rb
```diff
@@ -12,7 +12,6 @@ class AnalysisServicesServerUpdateParameters
 
       include MsRestAzure
 
-      include MsRest::JSONable
```
veronicagg: why was this included before?

sarangan12: It was a remnant from a serialization task which was later dropped.

sarangan12:
@veronicagg Added the link in the comment.

PR #1018, opened 2017-10-04
regenerating scheduler, relay, links and traffic manager

veronicagg:


veronicagg:
@sarangan12 I regenerated links updating the title via config.json, I haven't published the update to config.json yet. I believe the same problem happens with locks though, and we've shipped that gem already, so not sure we want to overwrite the title for that one. Let's see if we can identify that file somehow else, how about the file ending in _client.rb?

PR #1039, opened 2017-10-10
Fix for cleaning the multiple versions folders under each sdk

sarangan12:
The current clean_generated task deletes the profiles folder (and as a result, the common folder inside the profiles folder). This is not the expected scenario. The Rakefile code has been modified to delete only the multiple API versions folder and not touch the profiles folder.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -30,13 +30,22 @@ namespace :arm do
     end
   end
 
-  desc 'Delete ./lib/generated for each sdk'
+  desc 'Delete multiple version folders for each sdk'
   task :clean_generated do
-    each_gem do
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -30,14 +30,24 @@ namespace :arm do
...
+        if subdir.to_s == 'profiles'
+          next
+        end
+
+        folder_to_be_cleaned = File.expand_path("../management/#{gem}/lib/#{subdir}", __FILE__)
+        puts "Cleaning folder - #{folder_to_be_cleaned}"
+        FileUtils.rm_rf(folder_to_be_cleaned)
+      end
     end
+    Dir.chdir(File.expand_path('..', __FILE__))
```
veronicagg: Please add another task to clean up rake file to be consistent among tasks on how we're using the directories.

PR #1041, opened 2017-10-11
Profile fixes

sarangan12:
@veronicagg @vishrutshah @salameer Included the details in the email sent to you. Please review.


azure_sdk/lib/profile_azure_stack/modules/compute_profile_module.rb
```diff
@@ -181,34 +181,32 @@ class ComputeClass
...
-        @resource_skus = client.resource_skus
-        @virtual_machines = client.virtual_machines
-        @virtual_machine_scale_sets = client.virtual_machine_scale_sets
-        @virtual_machine_scale_set_extensions = client.virtual_machine_scale_set_extensions
-        @virtual_machine_scale_set_vms = client.virtual_machine_scale_set_vms
-        @disks = client.disks
-        @snapshots = client.snapshots
-        @virtual_machine_run_commands = client.virtual_machine_run_commands
+
+        client_0 = Azure::ARM::Compute::Api_2017_03_30::ComputeManagementClient.new(configurable.credentials, base_url, options)
```
vishrutshah: minor: Although this is internally used name `client_0` but it'd be great if we can have `client_2017_03_30` just for readability but not required as of today

azure_sdk/lib/profile_latest/profile_latest_profile_client.rb
```diff
@@ -81,62 +81,62 @@ def initialize(options = {})
...
+      @operational_insights = OperationalInsights::OperationalInsightsClass.new(self)
       @policy = Policy::PolicyClass.new(self)
-      @powerbiembedded = PowerBiEmbedded::PowerBiEmbeddedClass.new(self)
-      @recoveryservices = RecoveryServices::RecoveryServicesClass.new(self)
-      @recoveryservicesbackup = RecoveryServicesBackup::RecoveryServicesBackupClass.new(self)
-      @recoveryservicessiterecovery = RecoveryServicesSiteRecovery::RecoveryServicesSiteRecoveryClass.new(self)
+      @power_bi_embedded = PowerBiEmbedded::PowerBiEmbeddedClass.new(self)
+      @recovery_services = RecoveryServices::RecoveryServicesClass.new(self)
+      @recovery_services_backup = RecoveryServicesBackup::RecoveryServicesBackupClass.new(self)
+      @recovery_services_site_recovery = RecoveryServicesSiteRecovery::RecoveryServicesSiteRecoveryClass.new(self)
```
vishrutshah: 👍 👍

profile_specs/monitor_profiles.json
```diff
@@ -1,48 +1,11 @@
...
-      }
-    },
-    "output_dir": "../../../management/azure_mgmt_monitor/lib/profiles",
-    "individual_gem_profile": true
-  },
-  {
-    "name": "Profile_2017_05_01_Preview",
-    "resourceTypes": {
-      "Microsoft.Monitor" : {
+        "2015-04-01": ["*"],
```
veronicagg: if we need to pick specific specs, how we would do that? I think we could go with this approach for now, but it's actually not safe in my opinion, to include everything under an api-version, packages/tags may pick specific files.

profile_specs/profiles.json
```diff
@@ -21,6 +21,7 @@
         "2016-09-01": ["*"]
       },
       "Microsoft.Policy" : {
```
veronicagg: should this be pointing to the specific Policy profile, defined in the policy gem?
similarly for the rest for the rest of the entries in the file?
Is Azure Stack profile definition done at package level? spec level? (I'm asking about the source from where we are getting the info)

generators/profilegen/src/profile_generator.rb
```diff
@@ -82,16 +82,45 @@ def generate_modules
...
       file = get_module_file resource_provider
       file.write(get_renderer(ProfileTemplates.module_template))
       @model_types, @operation_types, @versions_clients_mapper = [], [], {}
+      @clients_ops_mapper = {}
       @management_client = ''
       @default_rp_client_version = ''
     end
   end
 
+  def check_available_after_index(operation_type_to_check, index_after_to_compare)
```
veronicagg: comments? 
we need to add comments throughout

generators/profilegen/src/profile_generator.rb
```diff
@@ -185,14 +214,16 @@ def get_client_file
     File.new(file_name, 'w')
   end
 
-  def check_for_availability(array_of_hash, key)
+  def check_and_delete_if_available(array_of_hash, key)
     array_of_hash.each do |hash_entry|
       if(hash_entry.keys.include?(:model_name))
         if(hash_entry[:model_name] == key)
+          array_of_hash.delete(hash_entry)
```
veronicagg: how do you know you're keeping the latest in the hash? (I believe you mentioned you would keep latest only)

generators/profilegen/src/resources/templates/module_template.template
```diff
@@ -22,42 +26,43 @@ module Azure::<%- if !@individual_gem_profile -%>Profiles::<%- end -%><%- if @in
...
-      # Method to get the client object based on the version
-      # If the version is invalid, an exception is raised
-      #
-      # @param version [String] The version of the required
-      # client object.
-      #
-      # @return Client object
-      #
+      <%- if @individual_gem_profile -%>
+      def get_client(version = '<%= @versions_clients_mapper.keys[@versions_clients_mapper.keys.length - 1] %>')
```
veronicagg: are we depending on order of the keys here to set up a default?

generators/profilegen/src/resources/templates/module_template.template
```diff
@@ -22,42 +26,43 @@ module Azure::<%- if !@individual_gem_profile -%>Profiles::<%- end -%><%- if @in
...
+      <%- else -%>
       def get_client(version = '<%= @default_rp_client_version %>')
```
veronicagg: where's the default_rp_client_version version getting picked up from? Should default_rp_client_version be assigned "@versions_clients_mapper.keys[@versions_clients_mapper.keys.length - 1]" for the case of individual gem profile, and so the same template code is applicable?

sarangan12:
Thanks @vishrutshah and @veronicagg We will definitely revisist the points post release. Since the code is approved now, merging it

Issue #976, opened 2017-09-13
[Release] Azure Stack related releases


vishrutshah:
- [x] Release Autorest.Ruby azure-arm generator for azure stack
- [ ] Release azure_mgmt_* gems with all the api-version included
- [ ] Release Profiles gem
   - [ ] Stack profile
   - [ ] Latest profile
- [x] Find the version of the gems to release 
- [x] Make sure to update changelog for with the SHA of the public swagger repo from where we generated things for tracking purpose

veronicagg:
- I've generated all the gems in PR https://github.com/Azure/azure-sdk-for-ruby/pull/925 waiting for confirmation on which ones to ship or not. 
- [] Need https://github.com/Azure/azure-sdk-for-ruby/issues/905 for each gem

veronicagg:
- [ ] Need to update release jobs
- [ ] Need to know if we'll have swagger_to_sdk working https://github.com/Azure/azure-sdk-for-ruby/issues/904
- [ ] Updates to readmes and docs

PR #1045, opened 2017-10-11
015 regen

veronicagg:
Updates:
- config.json file
    - added one more version of resources per reques
- version to 0.15.0
- cherry picked commit for 10 second delay on publishing from master
- regenerated code with autorest.ruby 2.0.19 + multi-api version changes from pending PR

Related: https://github.com/Azure/azure-sdk-for-ruby/issues/976


sarangan12:
@veronicagg I am ok with the first 2 commits. Just a QQ. Since we are changing the version here, should we bring in the last change log changes and update new change log also. 

Also, why is the build failing????

sarangan12:
I see the reason for failure. How will we handle it usually?

veronicagg:
I've updated version.rb files due to build failures. Problem is that version.rb file at package level won't get updated automatically by autorest, because it's not generated by it. 
We need to fix this too, though, we are expecting to rev packages independently, so we have to address that too.

sarangan12:
The changes at https://github.com/Azure/azure-sdk-for-ruby/pull/1009/commits/d52eba5e7ade689b0e9c98aa271d103fafcfffbe has not been brought in completely. The spec changes in this PR must also be brought for the build to succeed.

sarangan12:
You can use this one https://github.com/Azure/azure-sdk-for-ruby/pull/1009/commits

swagger_to_sdk_config.json
```diff
@@ -1,6 +1,6 @@
 {
   "meta": {
-    "version": "0.2.0",
+    "version": "0.3.0",
```
vishrutshah: This is the version of the swagger_to_sdk...it must be 0.2.0

vishrutshah: @sarangan12 has taken care for this one 👍 👍

vishrutshah:
Let's see why CI is not passing :)

sarangan12:
CI has passed. Planning to merge the changes

PR #1046, opened 2017-10-11
Profiles regenerated

sarangan12:
1. Renamed Azure Stack to Profile_2016_09_01
2. Modified the profile to be in consitent with https://github.com/Azure/azure-cli/blob/master/src/azure-cli-core/azure/cli/core/profiles/_shared.py#L27-L86
3, Regenerated all profiles & require files. Tested against the sample and the other two test programs I have and confirmed that they are working fine. 

@veronicagg @vishrutshah @salameer Please review and approve


profile_specs/profiles.json
```diff
@@ -1,15 +1,15 @@
 [
   {
-    "name": "Profile_Azure_Stack",
+    "name": "Profile_2016_09_01",
```
veronicagg: why did the definition change so much? just curious

Issue #1034, opened 2017-10-09
Post Profile Release Activity


sarangan12:
1. The Monitor SDK has odata.type which has a temporary patch. This must be removed and a proper solution must be found. 
2. Unification of meta-data
3. task to evaluate if we can simplify the each gem's profile metadata, we have a lot of files (one for each gem) at this point that would require manual update in multiple places with similar data.
4. Look at https://github.com/Azure/azure-sdk-for-ruby/pull/1050#pullrequestreview-69299444
5. Reduce the number of configuration files and consolidate them in one place.
6. Add more documentation in the Rakefile
7. Remove Updating versions task from regen multiple SDK
8. In Rakefile, try removing hardcoded azure_sdk
9. Try moving the hardcoded paths to constants and to a single location
10. Look at https://github.com/Azure/azure-sdk-for-ruby/pull/1064 and cover the feedback.

veronicagg:
@sarangan12 could you open an issue to list the remaining tasks from this issue and the feedback I left in the last PR that was merged? This issue contained a lot of things, and it's hard to know what's been addressed versus not, so I'd prefer that we split and close this one with just the set of things that were addressed.

veronicagg:
task tracking #4 above https://github.com/Azure/azure-sdk-for-ruby/issues/1074

PR #1050, opened 2017-10-12
Namespace changes

sarangan12:
@veronicagg @vishrutshah @devigned @salameer Per the latest discussion I have made the changes to the namespaces and tested it out. Please review and approve


sarangan12:
@devigned Done

profile_specs/analysis_services_profiles.json
```diff
@@ -1,16 +1,6 @@
...
-      "Microsoft.AnalysisServices" : {
-        "2016-05-16": ["*"]
-      }
-    },
-    "output_dir": "../../../management/azure_mgmt_analysis_services/lib/profiles",
-    "individual_gem_profile": true
-  },
-  {
-    "name": "Profile_2017_07_14",
+    "name": "V2017_07_14",
```
veronicagg: should this name be the date of our release? and maybe prefixed with something different than "V", what if a profile comes out afterwards and the name conflicts with ours? Maybe "Snapshot_2017_10_12", anything shorter would be great, thoughts?

sarangan12: I prefer V since it is clean, short and simple. But, I am curious about the name conflict. How could a future version conflict with something of past date????

veronicagg: @sarangan12 well, we don't know what date/name could be used for profiles that are not defined by us, for example, we're currently needing to use "v2017_03_09" as version for a "pre-defined" profile, so if sometime later, a profile gets defined with a date that coincides with one of our release dates, it would conflict.

veronicagg:
@sarangan12 in the last commit, it's hard to see the changes done manually, since it's all together, were there only profile metadata changes done manually? thanks.

sarangan12:
@veronicagg I assume you are mentioning https://github.com/Azure/azure-sdk-for-ruby/pull/1050/commits/d9aa1a501d99d1ca703a391c30d7801b9c5d1c1c as the last commit.

There are no manual changes to this commit. It is just merge from azure-stack and regenerated all.

veronicagg:
@sarangan12 looks good, thanks! We'll need to evaluate how we pick certain swagger files instead of everything in each folder afterwards.

veronicagg:
https://github.com/Azure/azure-sdk-for-ruby/issues/1034

PR #1051, opened 2017-10-13
Regen namespace updates

veronicagg:
- Updates to config.json and swagger_to_sdk for new namespaces
- Updates to version and module_definition files to match new namespaces
- Updates to tests (specs folder) to match updates
- Regenerated all gems based on new namespaces with @microsoft.azure/autorest.ruby@v3.0.20


veronicagg:
one more comment/question, graph is actually data plane, should we update the namespace to reflect that?

sarangan12:
Looks good. Note to self: get_version of profile generator must be changed 'after' your merge.

PR #1064, opened 2017-10-18
Multiple Upgrades to the Azure Ruby SDK

sarangan12:
**Combined all the profiles json files to one file**
There were 50+ files for profile specifications, each specific to a service. This will be difficult to maintain in the future. So, With this PR, all the 50+ files have been combined to one file (and corresponding code changes to handle the merge)

**Bug fix to include azure_sdk in build and release process**
The build and release process did not include the azure_sdk. As a result, this gem was not included automatically in the previous release and needed manual intervention. So, With this PR, that issue has been fixed and the azure_sdk has been included in the build and release process.

**Removed all the relative paths**
The Metadata files and profiles files have relative paths in them which would cause issues in the future (such as moving the generator out). With this PR, all the relative paths have been removed.

**Split the rollup and individual templates to get a cleaner code**
The Template files for the Rollup profiles and Individual profiles were same. So, they had several if conditions all over which would definitely cause issues in the future when we want to make changes. With this PR, they are completely split and independent which would make the future enhancements easier. 

**Cleaned the Rakefile**
The Rakefile had some unnecessary code and also complex Directory changes which will result in warnings during the regeneration. With this PR, such issues have been cleared. 

**None of these changes resulted in any code changes within the SDK**. I have executed the _bundle exec rake arm:regen && bundle exec rake arm:regen_all_profiles_ and this has caused **0** changes.

@veronicagg @vishrutshah @salameer @devigned Please review and approve


veronicagg:
@sarangan12 I believe you're addressing parts of https://github.com/Azure/azure-sdk-for-ruby/issues/1034. Could you split out an issue that contains only the things you planned to address with this PR? I'd like to make sure we're able to track what's pending and have issues with titles that represent what's getting addressed.

sarangan12:
Added a new commit . to include code changes related to Independent Versioning of gems. Provided a short description on the changes included:

* Deleted the old ARM_VERSION file. 
* Introduce 2 new files:
    * GEM_VERSIONS
    * GEMS_TO_RELEASE
* **What does the GEM_VERSIONS file do?**
   The GEM_VERSIONS file replaces the ARM_VERSION file. This file consists of all gems including their versions. This will be picked up by the regen task and the version value will be used as the value of the package-version attribute. 
   In addition, this file will be used by the gemspec file of the azure_sdk rollup gem. 
* **What does the GEMS_TO_RELEASE file do?**
   The GEMS_TO_RELEASE file consists of the gems to be released. This will be used only by the release task. 
* Other Miscellaneous points:
   * The regen tasks also updates the version file of the individual gems and also the version file of the azure_sdk rollup gem.
   * Added require file regeneration as an enhancement to the regen task. 

**How will these changes impact the Jenkins Publish Job?**
No significant changes. The only change will be to modify the required ARM_VERSION parameter to date to be used as the tag. I will do that change once this PR is approved.

generators/profilegen/src/profile_generator.rb
```diff
@@ -69,12 +68,9 @@ def generate_modules
       @spec_includes << @module_require
```
veronicagg: could there be more than module_required? 
could there be several RPs pointing to the same module_required?

generators/profilegen/src/profile_generator.rb
```diff
@@ -69,12 +68,9 @@ def generate_modules
...
       @class_name     = get_ruby_specific_resource_type_name(resource_provider)
       @class_names   << @class_name
-      if(!@individual_gem_profile)
```
veronicagg: if this file will only be used for individual gem profile, should its name be updated to indicate this?

generators/profilegen/src/resources/templates/rollup/client_template.template
```diff
@@ -2,45 +2,23 @@
...
-require 'profiles/<%= @profile_name.downcase %>/<%= @module_definition_file_name.gsub('.rb','') %>'
-<%- @file_names.each do |file_name| -%>
-require 'profiles/<%= @profile_name.downcase %>/modules/<%= file_name %>'
-<%- end -%>
-require 'profiles/common/configurable'
-require 'profiles/common/default'
-<%- else -%>
 require '<%= @profile_name.downcase %>/<%= @module_definition_file_name.gsub('.rb','') %>'
 <%- @file_names.each do |file_name| -%>
 require '<%= @profile_name.downcase %>/modules/<%= file_name %>'
```
veronicagg: similar comment to the one above here, it'd be great to simplify the template a bit, so it's more readable - anything that could help with that would be appreciated for future maintenance.

generators/profilegen/src/resources/templates/rollup/client_template.template
```diff
@@ -2,45 +2,23 @@
...
   #
-  <%- if @individual_gem_profile -%>
-  class Client < <%= "#{@class_names[0]}" %>Class
-  <%- else -%>
   class Client
-  <%- end -%>
     include Azure::ARM::Configurable
 
-    <%- if !@individual_gem_profile -%>
     attr_reader <% @class_names.each_with_index do |class_name, index| %> :<%= class_name.gsub(/([a-z\d])([A-Z])/,'\1_\2').downcase %><% if index != @class_names.length - 1 %>,<% end %><% end %>
```
veronicagg: similarly here, could we have less logic in the template and more in the code behind?

generators/profilegen/src/resources/templates/individual/module_template.template
```diff
@@ -4,14 +4,7 @@
...
 module Azure::<%= @class_name %>::Profiles::<%= @profile_name %>::Mgmt
-<%- else -%>
-module Azure::Profiles::<%= @profile_name %>
-<%- end -%>
-  <%- if !@individual_gem_profile -%>
-  module <%= @class_name %>::Mgmt
-  <%- end -%>
     <%- @clients_ops_mapper.each do |key, value| -%>
     <%- value.each do |operation_type| -%>
     <%- if(operation_type[:operation_name_ruby] != 'DO_NOT_ADD') -%>
```
veronicagg: can't remember what's the DO_NOT_ADD for?

generators/profilegen/src/profile_templates.rb
```diff
@@ -9,25 +9,38 @@
...
+  Profile_gen_resources_path = 'generators/profilegen/src/resources'
   #
   # Module template to generate the individual module
   # files per RP (such as ComputModule, NetworkModule, etc)
   #
-  def self.module_template
-    File.read(ProfileTemplateFiles::MODULE_TEMPLATE)
+  def self.module_template(sdk_path, individual)
+    if(individual)
+      File.read("#{sdk_path}/#{Profile_gen_resources_path}/#{ProfileTemplateFiles::Individual::MODULE_TEMPLATE}")
```
veronicagg: is there a reason to construct more pieces of the path here, instead of having them defined directly in the constant?

generators/profilegen/src/profile_template_files.rb
```diff
@@ -7,7 +7,15 @@
 # in the generation of profile SDKs
 #
 module ProfileTemplateFiles
-  MODULE_TEMPLATE = './resources/templates/module_template.template'
-  CLIENT_TEMPLATE = './resources/templates/client_template.template'
-  MODULE_DEFINITION_TEMPLATE = './resources/templates/module_definition_template.template'
+  module Rollup
```
veronicagg: are these constants only used in the profile_template.rb file? if so, should we define them directly there?

generators/profilegen/src/profile_generator_options_parser.rb
```diff
@@ -13,6 +13,9 @@
 #    -h, --help : Displays help for the profile generator
 #    -d, --dir_metadata : File that contains metadata info about RPs
 #    -p, --profile : Input file contains the profile generation info about RPs, versions, etc.
+#    -m, --mode : Mode can be 'rollup'/ 'management'
+#    -k, --key : Key of the Profile
```
veronicagg: what's key of the profile?

generators/profilegen/src/profile_generator_options_parser.rb
```diff
@@ -13,6 +13,9 @@
...
+#    -s, --sdk_path : Path to Azure Ruby SDK
```
veronicagg: can you provide more info on what the format of the path here should be?

generators/profilegen/src/profile_generator_options_parser.rb
```diff
@@ -13,6 +13,9 @@
 #    -h, --help : Displays help for the profile generator
 #    -d, --dir_metadata : File that contains metadata info about RPs
 #    -p, --profile : Input file contains the profile generation info about RPs, versions, etc.
+#    -m, --mode : Mode can be 'rollup'/ 'management'
```
veronicagg: would we not support "dataplane"? what would be different profile generator wise? is there a reason why the options are not rolloup & individual as in other cases?

generators/profilegen/src/profile_generator.rb
```diff
@@ -69,12 +68,9 @@ def generate_modules
...
-        @default_rp_client_version = @default_versions[resource_provider]
-      end
 
       resource_types_obj.each do |resource_type_version, resource_types|
```
veronicagg: could you add some inline comments to help follow the logic?

generators/profilegen/src/profile_generator.rb
```diff
@@ -126,7 +121,7 @@ def check_available_after_index(operation_type_to_check, index_after_to_compare)
   #
   def generate_client
     file = get_client_file
```
veronicagg: similar to a comment I made in another place, should we have a method to create a file based on a name and template, seems that we do that in multiple places and not very consistent across the generator methods.

generators/profilegen/src/resources/templates/individual/client_template.template
```diff
@@ -21,15 +21,5 @@ module Azure::<%= @class_name %>::Profiles::<%= @profile_name %>::Mgmt
       super(options)
     end
 
-    def credentials
-      if @credentials.nil?
```
veronicagg: shouldn't this change have come as part of the merge? with merged PR https://github.com/Azure/azure-sdk-for-ruby/pull/1090/files ?
I'm curious on what else we could have missed in the merge, therefore the question.

Rakefile
```diff
@@ -242,7 +247,7 @@ Rake::Task['arm:regen_individual_profiles'].enhance do
   Rake::Task['arm:regen_individual_require_files'].invoke
 end
 
-Rake::Task['arm:regen'].enhance do
+Rake::Task['arm:regen_sdk_versions'].enhance do
```
veronicagg: change is ok. I think I made a comment on this before, on what's the benefit of enhance? could we be consistent and have the regen_sdk_versions task just call regen_individual_require_files as a pre-req?

sarangan12: Actually, You cannot make regen_individual_require_files a pre requisite. If you execute regen_individual_require_files before regen_sdk_versions task, the require file will be empty. Because, in order to generate the require files, the sdk versions must be generated. i.e. you need to execute regen_individual_require_files task AFTER the regen_sdk_versions task. Now, it is possible to create a third task and make both of them as pre reqs. But, according to rake specification the order is not guaranteed. So, enhancement is the only option. Also, enhancement is standard design option provided by Rake and I believe we must start using it.

PR #1090, opened 2017-10-24
Credentials fix

sarangan12:
This PR is to fix the issue pointed in Stuart's email. Please review and approve. 
Addressing https://github.com/Azure/azure-sdk-for-ruby/issues/1092

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,41 @@ Note: x64 Ruby for Windows is known to have some compatibility issues.
 
 # Getting Started with Azure Resource Manager Usage (Preview)
 
+## Prerequisite
+
+In order to use the Azure SDK, you must supply the following values to the Azure SDK:
+
+* Tenant Id
+* Client Id
+* Subscription Id
+* Client Secret
+
+You could pass the above values in the following ways:
+
+### Option 1 - Environment Variables
+You can set the (above) values using the following environment variables:
+
+* AZURE_TENANT_ID
+* AZURE_CLIENT_ID
+* AZURE_SUBSCRIPTION_ID
+* AZURE_CLIENT_SECRET
+
+### Option 2 - Options Hash
+You can set the (above) values using the options hash:
+
+```ruby
+options = {
+  tenant_id: 'YOUR TENANT ID',
+  client_id: 'YOUR CLIENT ID',
+  client_secret: 'YOUR CLIENT SECRET',
+  subscription_id: 'YOUR SUBSCRIPTION ID'
+}
+```
+
+### Option 3 - Combination of Environment Variables & Options Hash
+You can set the (above) values using a combination of environment variables and options hash. The values mentioned in the options hash will take precedence over the environment variables.
+
+
 ## Install the rubygem packages
 
 You can install the azure rubygem packages directly.
@@ -218,8 +253,7 @@ purchase_plan_obj = Azure::Compute::Profiles::Latest::Mgmt::Models::PurchasePlan
 
 ## Usage of Individual gem using using specific api-version
 
-In the previous section, we used the profile associated with individual gem. In the current section, we could use the
-version directly.
+In the previous section, we used the profile associated with individual gem. In the current section, we could use the version directly.
 
 ### Install
 
@@ -234,16 +268,15 @@ The following lines should be used to instantiate a profile client:
 
 ```ruby
 # Provide credentials
-options = {
-  tenant_id: ENV['AZURE_TENANT_ID'],
-  client_id: ENV['AZURE_CLIENT_ID'],
-  client_secret: ENV['AZURE_CLIENT_SECRET'],
-  subscription_id: ENV['AZURE_SUBSCRIPTION_ID']
-}
+provider = MsRestAzure::ApplicationTokenProvider.new(
+       ENV['AZURE_TENANT_ID'],
+       ENV['AZURE_CLIENT_ID'],
+       ENV['AZURE_CLIENT_SECRET'])
+credentials = MsRest::TokenCredentials.new(provider)
 
 # Target client for 2016_03_30 version of Compute
 compute_client = Azure::Compute::Mgmt::V2016_03_30::ComputeManagementClient.new(credentials)
-compute_client.subscription_id = subscription_id
+compute_client.subscription_id = ENV['AZURE_SUBSCRIPTION_ID']
 ```
 
 The compute client could be used to access operations and models:
--- a/README.md
+++ b/README.md
@@ -82,7 +82,21 @@ You can set the (above) values using the following environment variables:
 * AZURE_SUBSCRIPTION_ID
 * AZURE_CLIENT_SECRET
 
+To set the environment variables, in Windows, you could use the command such as:
+
+```
+set AZURE_TENANT_ID=<YOUR_TENANT_ID>
+```
+
+In Unix based systems, you could use the command such as:
+
+```
+export AZURE_TENANT_ID=<YOUR_TENANT_ID>
+```
+
 ### Option 2 - Options Hash
+The initialization of profile clients take an options hash as a parameter. This options hash consists of tenant_id, client_id, client_secret, subscription_id, active_directory_settings and credentials. Among these, the active_directory_settings and credentials are optional.
+
 You can set the (above) values using the options hash:
 
 ```ruby
@@ -91,6 +105,24 @@ options = {
   client_id: 'YOUR CLIENT ID',
   client_secret: 'YOUR CLIENT SECRET',
   subscription_id: 'YOUR SUBSCRIPTION ID'
+}
+```
+
+If you would like to pass in the credentials object, you could use the the following code:
+
+```ruby
+provider = MsRestAzure::ApplicationTokenProvider.new(
+       ENV['AZURE_TENANT_ID'],
+       ENV['AZURE_CLIENT_ID'],
+       ENV['AZURE_CLIENT_SECRET'])
+credentials = MsRest::TokenCredentials.new(provider)
+
+options = {
+  tenant_id: 'YOUR TENANT ID',
+  client_id: 'YOUR CLIENT ID',
+  client_secret: 'YOUR CLIENT SECRET',
+  subscription_id: 'YOUR SUBSCRIPTION ID',
+  credentials: credentials
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -62,6 +62,39 @@ Additional info on Azure deployment models [https://azure.microsoft.com/en-us/do
 Note: x64 Ruby for Windows is known to have some compatibility issues.
 
 # Getting Started with Azure Resource Manager Usage (Preview)
+## Install the rubygem packages
+
+You can install the azure rubygem packages directly.
+
+```bash
+gem install azure_mgmt_compute
+gem install azure_mgmt_storage
+gem install azure_mgmt_resources
+gem install azure_mgmt_network
+```
+
+Or use them in your Gemfile.
+
+```Ruby
+gem 'azure_mgmt_storage'
+gem 'azure_mgmt_compute'
+gem 'azure_mgmt_resources'
+gem 'azure_mgmt_network'
+```
+
+Be aware the Azure Resource Manager Ruby SDK is in preview and will likely have breaking interface changes in upcoming
+releases. An increased number in Minor version may indicate breaking changes. 
+
+### Authentication
+
+The first step to using the SDK is authentication and permissions. For people unfamilar with Azure this may be one of
+the more difficult concepts. For a reference on setting up a service principal from the command line see
+[Authenticating a service principal with Azure Resource Manager](http://aka.ms/cli-service-principal) or
+[Unattended Authentication](http://aka.ms/auth-unattended). For a more robust explanation of authentication in Azure,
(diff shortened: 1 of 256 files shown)


generators/profilegen/src/resources/common/configurable.rb
```diff
@@ -0,0 +1,84 @@
...
+    # Resets the configurable options to provided options or defaults.
+    # This will also creates MsRest::TokenCredentials to be used for subsequent Azure Resource Manager clients.
+    #
+    def reset!(options = {})
+      Azure::Common::Configurable.keys.each do |key|
+        default_value = Azure::Common::Default.options[key]
+        instance_variable_set(:"@#{key}", options.fetch(key, default_value))
+      end
+
+      default_value = get_credentials(self.tenant_id, self.client_id, self.client_secret, self.active_directory_settings)
```
veronicagg: is this needed? 
if credentials are not passed in the options, should the "fetch" above get "credentials" from Configurable and set the value to the same thing that "get_credentials" below is doing?

veronicagg: I wonder why we need to get credentials in Configurable and remove it from Default, where it was.

README.md
```diff
@@ -195,14 +191,10 @@ The following lines should be used to instantiate a profile client:
...
-       ENV['AZURE_CLIENT_SECRET'])
-credentials = MsRest::TokenCredentials.new(provider)
-
 options = {
-  credentials: credentials,
+  tenant_id: ENV['AZURE_TENANT_ID'],
+  client_id: ENV['AZURE_CLIENT_ID'],
+  client_secret: ENV['AZURE_CLIENT_SECRET'],
   subscription_id: ENV['AZURE_SUBSCRIPTION_ID']
 }
```
veronicagg: would our guidelines from before profile changes not work for initializing credentials, I believe we were doing:
```ruby
 provider = MsRestAzure::ApplicationTokenProvider.new(
        ENV['AZURE_TENANT_ID'],
        ENV['AZURE_CLIENT_ID'],
        ENV['AZURE_CLIENT_SECRET'])
    credentials = MsRest::TokenCredentials.new(provider)
    @client = Azure::ARM::Resources::ResourceManagementClient.new(credentials)
    @client.subscription_id = @subscription_id
```
If so, should we leave them as they were, just updating the namespace for ResourceManagementClient? so we demonstrate the minimum change?

veronicagg: BTW, after this PR is done, we'd need to re-update the samples that were updated with the last release too.

sarangan12: The detailed explanation would be that we always want the user (of the SDK) to provide us with the tenant_id, client_id, client_secret and subscription_id. There is no escaping that. The credentials is just a derived value from these. The user may or may not provide it. 

But, when you think about it, why would a user want to provide a derived value when he/she is supplying the original values anyway? That is the reason, I am updating the examples in readme and removed the credentials. But, if a user wants to provide it, then he is free to do that.

generators/profilegen/src/resources/common/configurable.rb
sarangan12: As mentioned in the previous comment, the user may or may not provide credentials in the options. If he provides it, we will use it. Else, we will derive it from the other values supplied. 

The 2 fetch are different. The first one (in setup_options) is getting the credentials from the default value and the second one is from the user supplied values (not the environment variables).

generators/profilegen/src/resources/common/configurable.rb
```diff
@@ -0,0 +1,84 @@
...
+    def config
+      self
+    end
+
+    private
+
+    #
+    # configures configurable options to default values
+    #
+    def setup_options
```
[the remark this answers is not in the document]
sarangan12: Sure. WIll change the name and regen in a few minutes

sarangan12:
@veronicagg Changed the name of the method and regened everything. Please review and approve

README.md
devigned: So... Credentials are not a bad thing to provide. This abstraction allows any token provider to be used.

sarangan12:
@veronicagg Modified per offline discussion

README.md
```diff
@@ -63,6 +63,41 @@ Note: x64 Ruby for Windows is known to have some compatibility issues.
...
+
+You could pass the above values in the following ways:
+
+### Option 1 - Environment Variables
+You can set the (above) values using the following environment variables:
+
+* AZURE_TENANT_ID
+* AZURE_CLIENT_ID
+* AZURE_SUBSCRIPTION_ID
+* AZURE_CLIENT_SECRET
```
veronicagg: can we put simple instructions on how to set them in mac and windows? (I think we had this before)

README.md
```diff
@@ -63,6 +63,41 @@ Note: x64 Ruby for Windows is known to have some compatibility issues.
...
+
+### Option 2 - Options Hash
+You can set the (above) values using the options hash:
```
veronicagg: I think we need to give context on what the options hash is - for example "the options hash getting passed as parameter when constructing a client for the service/profile"

README.md
```diff
@@ -63,6 +63,41 @@ Note: x64 Ruby for Windows is known to have some compatibility issues.
 
 # Getting Started with Azure Resource Manager Usage (Preview)
 
+## Prerequisite
```
veronicagg: I think would place this section after the "Install the rubygem packages" section, as that one provides context on authentication. Then we can talk about how to set this for usage in the packages.'

README.md
```diff
@@ -63,6 +63,41 @@ Note: x64 Ruby for Windows is known to have some compatibility issues.
...
+```ruby
+options = {
+  tenant_id: 'YOUR TENANT ID',
+  client_id: 'YOUR CLIENT ID',
+  client_secret: 'YOUR CLIENT SECRET',
+  subscription_id: 'YOUR SUBSCRIPTION ID'
+}
+```
+
+### Option 3 - Combination of Environment Variables & Options Hash
```
veronicagg: what's the recommendation for the case in which the user wants to create credentials themselves? either from TokenProvider or MSITokenProvider? if they were to pass them in the hash, they would still need to pass in the 4 values specified above? so if credentials are passed, there are 5 values required for the hash?

generators/profilegen/src/resources/common/configurable.rb
```diff
@@ -50,7 +50,16 @@ def reset!(options = {})
         instance_variable_set(:"@#{key}", options.fetch(key, default_value))
       end
 
-      default_value = get_credentials(self.tenant_id, self.client_id, self.client_secret, self.active_directory_settings)
+      fail ArgumentError, 'tenant_id is nil' if self.tenant_id.nil?
```
veronicagg: can we give an indication on how to fix it? set the env vars or pass it in options?

veronicagg: also maybe match the message from https://github.com/Azure/azure-sdk-for-ruby/blob/ab9cdb6ef0f53741d76caab9c9e598f9781ca320/runtime/ms_rest_azure/lib/ms_rest_azure/credentials/application_token_provider.rb#L50

generators/profilegen/src/resources/common/configurable.rb
```diff
@@ -50,7 +50,16 @@ def reset!(options = {})
...
+      fail ArgumentError, 'client_id is nil' if self.client_id.nil?
+      fail ArgumentError, 'client_secret is nil' if self.client_secret.nil?
+      fail ArgumentError, 'subscription_id is nil' if self.subscription_id.nil?
+      fail ArgumentError, 'active_directory_settings is nil' if self.active_directory_settings.nil?
```
veronicagg: could this be nil at this point? this is not something we're asking the user to pass in, right?

README.md
sarangan12: Done

README.md
sarangan12: Done

README.md
sarangan12: Done

generators/profilegen/src/resources/common/configurable.rb
sarangan12: Technically, yes. They will not be nil. I just wanted to be cautious that one of our users intentionally passing nil value for active_directory_settings in the options hash.

stuartpreston:
I had a look at this and it seems that whilst we can now use our own credentials rather than environment variables to connect, somewhere along the way we've pushed the ApplicationTokenProvider out of the way.  Instead you now require us to send in the credentials and the settings directly to the client, whereas before we simply created an ApplicationTokenProvider with for the correct cloud environment and passed that in.  Is this change in approach by design? 

/cc @devigned @veronicagg @russellseymour

sarangan12:
@stuartpreston 

Thanks for providing the feedback. I would like to request some clarifications. You have mentioned that we have pushed the ApplicationTokenProvider out of the way. Before the release of the profiles, we accessed the individual gems, say azure_mgmt_resources like this:


```ruby
subscription_id = ENV['AZURE_SUBSCRIPTION_ID'] 
  provider = MsRestAzure::ApplicationTokenProvider.new(
      ENV['AZURE_TENANT_ID'],
      ENV['AZURE_CLIENT_ID'],
      ENV['AZURE_CLIENT_SECRET'])
credentials = MsRest::TokenCredentials.new(provider)
client = Azure::ARM::Resources::ResourceManagementClient.new(credentials)
client.subscription_id = subscription_id
```

So, we always passed the credentials to the client object. With the profiles, we could initialize the client as:

```ruby
subscription_id = ENV['AZURE_SUBSCRIPTION_ID']
provider = MsRestAzure::ApplicationTokenProvider.new(
      ENV['AZURE_TENANT_ID'],
      ENV['AZURE_CLIENT_ID'],
      ENV['AZURE_CLIENT_SECRET'])
credentials = MsRest::TokenCredentials.new(provider)

options = {
      credentials: credentials,
      subscription_id: subscription_id
}

client = Azure::Resources::Profiles::Latest::Mgmt::Client.new(options)
```

So, both before and after, we have always used credentials and not the ApplicationTokenProvider. I would like to know if we are missing any scenarios. 

If possible, I would like to setup a call where we could discuss your questions in detail. Please let me know.

stuartpreston:
Ok, so our new symptoms are that if I use your code but use implicit values rather than references to environment variables, the new Client does not recognize the credentials passed through in the credentials object.  Instead, we must set them explicitly:

```ruby
require 'azure_mgmt_resources'

subscription_id = '1e0b427a-xxmy-subs-crip-ee558463ebbf'
provider = MsRestAzure::ApplicationTokenProvider.new(
      'a2b2d6bc-xxmy-tena-ntid-f97a7ac416d7',
      '020fb5ca-xxmy-clie-ntid-40ac3b94b40a',
      'my-client-secret')
credentials = MsRest::TokenCredentials.new(provider)

options = {
      credentials: credentials,
      subscription_id: subscription_id
}

client = Azure::Resources::Profiles::Latest::Mgmt::Client.new(options)
```

```ruby
C:\projects\other\azuresdk015> chef exec ruby test2.rb
C:/Users/spreston/AppData/Local/chefdk/gem/ruby/2.4.0/gems/azure_mgmt_resources-0.15.1/lib/profil
es/common/configurable.rb:53:in `reset!': tenant_id is nil (ArgumentError)
        from C:/Users/spreston/AppData/Local/chefdk/gem/ruby/2.4.0/gems/azure_mgmt_resources-0.15
.1/lib/profiles/latest/modules/resources_profile_module.rb:78:in `initialize'
        from C:/Users/spreston/AppData/Local/chefdk/gem/ruby/2.4.0/gems/azure_mgmt_resources-0.15
.1/lib/profiles/latest/resources_latest_profile_client.rb:19:in `initialize'
        from test2.rb:15:in `new'
        from test2.rb:15:in `<main>'
C:\projects\other\azuresdk015>
```

So it appears the defaults are still loaded from the environment variables if credentials are not passed.  Hence my comment that we can only use the client if we pass credentials (tenant_id, client_id, client_secret) directly and not through the TokenCredentials object, which leads to the question about why credentials passed along from a TokenCredentials provider seem to be ignored.

I am travelling this week so we'll have to stay asynchronous on this, next week looks a bit clearer in my calendar.

veronicagg:
@stuartpreston at the moment we require that the `options` contain values for tenant_id, client_id, and client_secret. If you're providing credentials yourself, that's good too, you'd add credentials key,value pair as well in the `options` (as you have above). We should be picking up the credentials you pass in this case, and only rely on the rest of the values to calculate default credentials. We should be making a change to avoid having to pass the tenant_id, client_id, client_secret, when credentials are passed in, then your code should work as is (https://github.com/Azure/azure-sdk-for-ruby/issues/1129)
Please see azure_mgmt_resources readme specifically on current usage https://github.com/Azure/azure-sdk-for-ruby/tree/master/management/azure_mgmt_resources#option-1---using-the-resources-profiles 
Would this work for you?

stuartpreston:
@veronicagg This is ok, but my thinking was that would want to abstract away from receiving credentials directly so that different auth methods can be used in future (e.g. we want to use a Certificate rather than a Client Secret).  Otherwise the purpose of TokenCredentials isn't that clear any more.

As long as we can specify credentials (and not only have them inferred from the environment), we're OK - our problem now is keeping up with these changes in the SDK.  The change from 0.14 to 0.15 was painful for us.

PS: Is there a way of using the certificate to authenticate today in the SDK? (as per https://gist.github.com/devigned/dae74a7ca54000f7b714)

PR #1095, opened 2017-10-27
Release 0.15.1

sarangan12:
@veronicagg @salameer This is to release a new version of azure sdk for ruby. Please review and approve


ChangeLog.md
```diff
@@ -1,3 +1,9 @@
+## 2017.10.30 - Azure ARM version 0.15.1
+
+* Regened all azure_mgmt_* sdks using [Autorest.2.0.4168](https://www.npmjs.com/package/autorest)(core) and [Autorest.Ruby.3.0.20](https://www.npmjs.com/package/@microsoft.azure/autorest.ruby) for release of ARM version 0.15.1
+* Fixed the [Issue #1092](https://github.com/Azure/azure-sdk-for-ruby/issues/1092) of passing tenant id, client id, client secret etc. Refer [PR #1090](https://github.com/Azure/azure-sdk-for-ruby/pull/1090) for further details.
```
veronicagg: I wonder how much we should call this a fix, yes, we're making changes, but I believe it's more about guidance on docs and returning an error than actually fixing a problem, thoughts? Wouldn't the current released code work if our recommendation on how to use it changes?

sarangan12: Actually no. If we just change the recommendation of including the tenant_id etc it options, the current release code will not work. Anyway, Changed 'Fixed' to 'Code changes related to'

sarangan12:
@veronicagg Yes. I missed the readme files inside the gems. With the latest commit, I have fixed it.

veronicagg:
@sarangan12 will you be closing all automatic PRs that came in since last release till now? As this PR includes regenerating everything, I believe those would need to be closed. Thanks!

Issue #1012, opened 2017-10-02
Should prevent creating VM with name longer than 15 characters


axelgMS:
Looks like Ruby SDK allows creating VM with a name longer than 15 characters. This should not be allowed.
Portal prevents this, and PowerShell too. eg:

$vmName = "TestResourceNameXYZ-TestDataDisk-TestResourceNames-ABCDEF-123456"

PS C:\> New-AzureRmVM -ResourceGroupName $resourceGroup -Location $location -VM $vmConfig
WARNING: Since the VM is created using premium storage, existing standard storage account, axelgsa47167195, is used for boot diagnostics.
**New-AzureRmVM : Windows computer name cannot be more than 15 characters long, be entirely numeric, or contain the following characters: ` ~ ! @ # $ % ^ & * ( ) = + _ [ ] { } \ | ; : . ' " , < > / ?.**
ErrorCode: InvalidParameter
ErrorMessage: Windows computer name cannot be more than 15 characters long, be entirely numeric, or contain the following characters: ` ~ ! @ # $ % ^ & * ( ) = + _ [ ] { } \ | ; : . ' " , < > / ?.
StatusCode: 400
ReasonPhrase: Bad Request
OperationID : 3ea56e6d-b0e2-4d18-98f9-6488bebd9903
At line:1 char:1
+ New-AzureRmVM -ResourceGroupName $resourceGroup -Location $location - ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : CloseError: (:) [New-AzureRmVM], ComputeCloudException
    + FullyQualifiedErrorId : Microsoft.Azure.Commands.Compute.NewAzureVMCommand

veronicagg:
thanks for reporting, we'll investigate and get back on this.

sarangan12:
@axelgMS Adding such constraints on the SDK depends on the entries in swagger. I have done necessary changes to Ruby SDK to handle the constraints if they are available. But, only the compute team could add such constraints to the swagger and I have created an issue https://github.com/Azure/azure-rest-api-specs/issues/1990 for the compute team. 

Since there is no pending action items, I will be closing this issue

PR #1119, opened 2017-11-08
Check for constraints before serialization

sarangan12:
This PR is to validate the constraints of the models before serialization. This is related to https://github.com/Azure/azure-sdk-for-ruby/issues/1012 & https://github.com/Azure/autorest/issues/2100

I am following a similar logic to node. https://github.com/Azure/azure-sdk-for-node/blob/master/runtime/ms-rest/lib/serialization.js#L72

@salameer @devigned @veronicagg Please review and approve


sarangan12:
Adding @amarzavery also for review as the code logic is taken from Node SDK.

PR #1127, opened 2017-11-13
Added metadata to gemspecs

dsexton:
See: http://guides.rubygems.org/specification-reference/#metadata

### DIFF
--- a/azure_sdk/azure_sdk.gemspec
+++ b/azure_sdk/azure_sdk.gemspec
@@ -19,6 +19,14 @@ Gem::Specification.new do |spec|
   spec.summary       = 'Official Ruby client library to consume Microsoft Azure services.'
   spec.homepage      = 'https://aka.ms/azure-sdk-for-ruby'
   spec.license       = 'MIT'
+  spec.metadata      = {
+    'bug_tracker_uri' => 'https://github.com/Azure/azure-sdk-for-ruby/issues',
+    'changelog_uri' => 'https://github.com/Azure/azure-sdk-for-ruby/blob/master/ChangeLog.md',
+    'documentation_uri' => 'https://azure.microsoft.com/en-us/develop/ruby/',
+    'homepage_uri' => 'https://aka.ms/azure-sdk-for-ruby',
+    'source_code_uri' => "https://github.com/Azure/azure-sdk-for-ruby/tree/ARM-#{Azure::VERSION}",
+    'wiki_uri' => 'https://github.com/Azure/azure-sdk-for-ruby/wiki'
+  }
 
   spec.files         = Dir["LICENSE.txt", "lib/**/*"]
   spec.bindir        = 'bin'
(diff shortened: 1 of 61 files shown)


azure_sdk/azure_sdk.gemspec
```diff
@@ -19,6 +19,14 @@ Gem::Specification.new do |spec|
   spec.summary       = 'Official Ruby client library to consume Microsoft Azure services.'
   spec.homepage      = 'https://aka.ms/azure-sdk-for-ruby'
   spec.license       = 'MIT'
+  spec.metadata      = {
+    'bug_tracker_uri' => 'https://github.com/Azure/azure-sdk-for-ruby/issues',
+    'changelog_uri' => 'https://github.com/Azure/azure-sdk-for-ruby/blob/master/ChangeLog.md',
+    'documentation_uri' => 'https://azure.microsoft.com/en-us/develop/ruby/',
+    'homepage_uri' => 'https://aka.ms/azure-sdk-for-ruby',
+    'source_code_uri' => "https://github.com/Azure/azure-sdk-for-ruby/tree/azure-v#{Azure::VERSION}",
```
veronicagg: I think you'd want this url to look like https://github.com/Azure/azure-sdk-for-ruby/tree/ARM-0.15.0 - notice "ARM" in the tag name.

management/azure_mgmt_analysis_services/azure_mgmt_analysis_services.gemspec
```diff
@@ -17,6 +17,13 @@ Gem::Specification.new do |spec|
   spec.summary       = 'Official Ruby client library to consume Microsoft Azure Analysis services.'
   spec.homepage      = 'https://aka.ms/azure-sdk-for-ruby'
   spec.license       = 'MIT'
+  spec.metadata      = {
+    'bug_tracker_uri' => 'https://github.com/Azure/azure-sdk-for-ruby/issues',
+    'changelog_uri' => 'https://github.com/Azure/azure-sdk-for-ruby/blob/master/ChangeLog.md',
+    'documentation_uri' => 'https://azure.microsoft.com/en-us/develop/ruby/',
+    'homepage_uri' => 'https://aka.ms/azure-sdk-for-ruby',
+    'source_code_uri' => 'https://github.com/Azure/azure-sdk-for-ruby/tree/master/management/azure_mgmt_analysis_services'
```
veronicagg: you could reference the ARM<version> tag here as well, if you'd like to point to the version, and pick up the version from https://github.com/Azure/azure-sdk-for-ruby/blob/master/management/azure_mgmt_analysis_services/lib/version.rb#L6

azure_sdk/azure_sdk.gemspec
dsexton: Thanks!

management/azure_mgmt_analysis_services/azure_mgmt_analysis_services.gemspec
dsexton: @veronicagg would there ever be a situation where the version of a given management spec diverges from the ARM version? If a bug is fixed in one of the management gems, do all the specs needs to be bumped to align?

Another possible option for the management specs is to do an abbreviated rev parse during build and include that to build a correlation between published gem version and source. That way the versions of these gems will be able to diverge, and you don't have to clutter the repository with git tags.

Let me know which route you prefer 😄

veronicagg: @dsexton yes, we will be releasing gems with different versions, so the versions may differ. We haven't thought much about the tags, so you're right, we may end up tagging with the name of the gem, not sure yet. Feel free to keep it as you have it, and we can update it later as we figure out the tag story.

PR #1158, opened 2017-12-01
Graph conversion

sarangan12:


Gemfile
```diff
@@ -69,4 +68,5 @@ gem 'azure_mgmt_stream_analytics',          path: 'management/azure_mgmt_stream_
 gem 'azure_mgmt_subscriptions',             path: 'management/azure_mgmt_subscriptions'
 gem 'azure_mgmt_traffic_manager',           path: 'management/azure_mgmt_traffic_manager'
 gem 'azure_mgmt_web',                       path: 'management/azure_mgmt_web'
+gem 'azure_graph',                          path: 'data/azure_graph'
```
veronicagg: should we call this gem graph_rbac? I'm wondering because there was another mention of Graph API (referring to Cosmos DB when we talked about samples) that was different recently.

Rakefile
```diff
@@ -50,6 +50,23 @@ namespace :arm do
         FileUtils.rm_rf(folder_to_be_cleaned)
       end
     end
+
+    Dir.chdir(File.expand_path('../data', __FILE__))
```
veronicagg: could we create a method we call instead of repeating the code from above and passing "data" folder instead of "management"?

Rakefile
```diff
@@ -70,6 +87,9 @@ namespace :arm do
         elsif(key == 'management')
           Dir.chdir("#{__dir__}/management/#{gem}")
           version = gem_versions['management'][gem]
+        elsif(key == 'data')
```
veronicagg: similar comment as above, could we parameterize this code?

Rakefile
```diff
@@ -286,9 +323,20 @@ def each_gem_dir
 
     yield subdir
   end
+
+  Dir.chdir("#{__dir__}/data")
```
veronicagg: similar to above comments, could we parameterize this code instead of copy/pasting?

Rakefile
```diff
@@ -300,6 +348,16 @@ def each_child
     end
   end
 
+  # Process Data Gems
```
veronicagg: same here

config.json
```diff
@@ -1,792 +1,791 @@
 {
   "azure_mgmt_analysis_services": {
     "azure_mgmt_analysis_services_2017_07_14": {
-      "markdown": "https://raw.githubusercontent.com/Azure/azure-rest-api-specs/7aa3a5247895ba34d6cfec73e036bb66dc907d20/specification/analysisservices/resource-manager/readme.md",
+      "markdown": "https://raw.githubusercontent.com/Azure/azure-rest-api-specs/4a97039db7d2ca3acd5a11fe32d815864bdfc27d/specification/analysisservices/resource-manager/readme.md",
```
veronicagg: are you just updating to the latest commit hash at the time you made this change?

data/azure_graph/azure_graph.gemspec
```diff
@@ -0,0 +1,35 @@
...
+
+require '../azure_graph/lib/module_definition'
+require '../azure_graph/lib/version'
+
+Gem::Specification.new do |spec|
+  spec.name          = 'azure_graph'
+  spec.version       = Azure::Graph::VERSION
+  spec.authors       = 'Microsoft Corporation'
+  spec.email         = 'azrubyteam@microsoft.com'
+  spec.description   = 'Microsoft Azure Active Directory Graph Data Client Library for Ruby'
```
veronicagg: Not sure we want to say "Data" here, you may just leave it out.

data/azure_graph/spec/1.6/graph_spec.rb
```diff
@@ -0,0 +1,49 @@
+# encoding: utf-8
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+
+require_relative 'spec_helper'
+
+describe 'GraphManagementClient' do
```
veronicagg: left over Management in the describe

generators/profilegen/src/defs/definition.rb
```diff
@@ -0,0 +1,24 @@
+# encoding: utf-8
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+
```
veronicagg: could you provide more context/explanation on the files you're adding for profile generation?

profile_specs/profiles.json
```diff
@@ -234,7 +366,9 @@
             "name": "Latest",
             "resourceTypes": {
                 "Microsoft.AnalysisServices": {
-                    "2017-07-14": ["*"]
+                    "management": {
```
veronicagg: why do we need a "management" tag here, if we already in the "management" section https://github.com/Azure/azure-sdk-for-ruby/pull/1158/commits/a818cec53fac531fce38474119376a72da2ff63b#diff-a886d4e0a33d2455b5a5344309de4389R363 ?

sarangan12:
Because of the recent merge this PR has got too many conflicts. I have opened a new PR https://github.com/Azure/azure-sdk-for-ruby/pull/1168 for this issue and addressed all your comments there

PR #1186, opened 2017-12-14
Preparing release for new gem azure_mgmt_msi

veronicagg:
Per the new spec https://github.com/Azure/azure-rest-api-specs/tree/13291dd9244584a66de3c53cc1a85998efd288b3/specification/msi/resource-manager
This PR is to release the Ruby package for msi service.


GEMS_TO_RELEASE
```diff
@@ -1,5 +1,5 @@
 {
     "management": [
-        "azure_mgmt_storage"
+        "azure_mgmt_msi"
```
yugangw-msft: Just a question, is this expected?

veronicagg: correct, I'm thinking I'll release this gem tomorrow, then do the updates to compute, since this one seems independent, right?

yugangw-msft: Yep, it is independent

Issue #1073, opened 2017-10-18
If a swagger spec is changed, have a task to generate the specific gem and related files


veronicagg:
Have a task/way of generating a service gem and necessary other related files (if anything needed in azure_sdk as well), when a spec gets updated. 
Evaluate scenario, where we'd want to ship a gem based on updates to specs that are included in it.

veronicagg:
split out of https://github.com/Azure/azure-sdk-for-ruby/issues/908

veronicagg:
Update this taks please, it seems that this is already to be covered under another affort or Task

sarangan12:
Task completed

PR #1223, opened 2018-01-09
Rakefile parameterization

sarangan12:
This is related to issue https://github.com/Azure/azure-sdk-for-ruby/issues/1073. 

What does this task do?
1. Parameterize the Rakefile
2. Provided a parameter to regen (such as azure_mgmt_compute) this would regenerate sdk versions and profile for that specific gem and the rollup.
3. If there is no parameter, then the current behavior will continue (that is every gem will be considered)

@veronicagg @salameer Please review and approve

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -130,9 +130,9 @@ REGEN_EXCLUDES = ['azure_sdk']
 desc 'Azure Resource Manager related tasks which often traverse each of the arm gems'
 namespace :arm do
   desc 'Delete ./pkg for each of the Azure Resource Manager projects'
-  task :clean, [:key] do |_, args|
+  task :clean, [:gem_name] do |_, args|
     each_gem do |dir|
-      if !should_i_continue(args[:key], dir)
+      if !should_i_continue(args[:gem_name], dir)
         next
       end
 
@@ -141,16 +141,16 @@ namespace :arm do
   end
 
   desc 'Delete multiple version folders for each sdk'
-  task :clean_generated, [:key] do |_, args|
-    clean_generated('management', args[:key])
-    clean_generated('data', args[:key])
+  task :clean_generated, [:gem_name] do |_, args|
+    clean_generated('management', args[:gem_name])
+    clean_generated('data', args[:gem_name])
     Dir.chdir(File.expand_path('..', __FILE__))
   end
 
   desc 'Build gems for each of the Azure Resource Manager projects'
-  task :build, [:key] => :clean do |_, args|
+  task :build, [:gem_name] => :clean do |_, args|
     each_gem do |dir|
-      if !should_i_continue(args[:key], dir)
+      if !should_i_continue(args[:gem_name], dir)
         next
       end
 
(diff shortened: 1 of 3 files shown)


Rakefile
```diff
@@ -130,22 +130,30 @@ REGEN_EXCLUDES = ['azure_sdk']
 desc 'Azure Resource Manager related tasks which often traverse each of the arm gems'
 namespace :arm do
   desc 'Delete ./pkg for each of the Azure Resource Manager projects'
-  task :clean do
-    each_gem do
+  task :clean, [:key] do |_, args|
+    each_gem do |dir|
+      if !should_i_continue(args[:key], dir)
```
veronicagg: should we be centralizing this logic more, like applying filtering at "each_child" method level?
 https://github.com/sarangan12/azure-sdk-for-ruby/blob/021d8322512ac3fe0a835cbb01d025d68e298bde/Rakefile#L450 and have a file that allows us to control which gems to generate/select? 
Having a parameter like it's added in this PR, is ok for individual gem generation and it's good, but I'm thinking if we should think a bit broader enabling generation of multiple gems too, so it facilitates our own steps. We could have a file similar to the "GEMS_to_release" which has a checked in sample with all the gem list, but with the purpose of editing locally for generation. Also, regarding "GEMS_to_release" file, I think we should enable our tasks to consume it, so that when we indicate a few gems to get released, the jobs only generate and build those gems (instead of doing everything and then skip the release step).

Rakefile
```diff
@@ -332,16 +362,24 @@ Rake::Task['arm:regen_rollup_profile'].enhance do
...
-Rake::Task['arm:regen_sdk_versions'].enhance do
-  Rake::Task['arm:regen_individual_require_files'].invoke
+Rake::Task['arm:regen_sdk_versions'].enhance do |_, args|
+  Rake::Task['arm:regen_individual_require_files'].invoke(args[:key])
 end
 
 task :default => :spec
 
+def should_i_continue(key, dir)
+  if(key.nil? || key == dir)
```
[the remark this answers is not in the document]
sarangan12: Done

Rakefile
```diff
@@ -130,22 +130,30 @@ REGEN_EXCLUDES = ['azure_sdk']
 desc 'Azure Resource Manager related tasks which often traverse each of the arm gems'
 namespace :arm do
   desc 'Delete ./pkg for each of the Azure Resource Manager projects'
-  task :clean do
-    each_gem do
+  task :clean, [:key] do |_, args|
```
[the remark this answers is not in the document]
sarangan12: Done

Rakefile
sarangan12: I think adding GEMS_TO_RELEASE to this would be overkill. It could be easily achieved by using the && command. Let us see if we really need that feature and then think about adding it.

veronicagg: Ok, as per my comment above I wouldn't tie it with GEMS_TO_RELEASE only, that comment was more of a convenience thing, as I think it can save some time (not building all gems), so it'd be good to consider. 
Regarding not modifying the approach right now, I'll leave it up to you, design wise I prefer the idea of centralizing logic and selecting from the start, instead of having code that skips gems repeated in multiple tasks.

Issue #1120, opened 2017-11-09
Add telemetry to profile clients


veronicagg:
We're currently adding info to the user agent for telemetry in each api-version client see https://github.com/Azure/azure-sdk-for-ruby/blob/master/management/azure_mgmt_compute/lib/2017-03-30/generated/azure_mgmt_compute/compute_management_client.rb#L185

We should probably add more to the user agent hen a profile is being used (to know from telemetry if profiles are being used) somewhere around;
https://github.com/Azure/azure-sdk-for-ruby/blob/master/generators/profilegen/src/resources/templates/individual/module_template.template#L51
and similarly for the roll-up profile.

sarangan12:
Task completed

PR #1229, opened 2018-01-09
Profile telemetry

sarangan12:
It is related to https://github.com/Azure/azure-sdk-for-ruby/issues/1120. 

Added code changes to add telemetry to the profiles.

@veronicagg @salameer Please review and approve.

### DIFF
--- a/azure_sdk/lib/latest/modules/analysisservices_profile_module.rb
+++ b/azure_sdk/lib/latest/modules/analysisservices_profile_module.rb
@@ -46,8 +46,7 @@ module Azure::Profiles::Latest
         end
 
         def add_telemetry(client)
-          profile_information = 'Profiles/Rollup/Mgmt/AnalysisServices'
-          profile_information = "#{profile_information}/Latest"
+          profile_information = 'Profiles/azure_sdk/Latest/AnalysisServices/Mgmt'
           client.add_user_agent_information(profile_information)
         end
 
--- a/azure_sdk/lib/latest/modules/analysisservices_profile_module.rb
+++ b/azure_sdk/lib/latest/modules/analysisservices_profile_module.rb
@@ -39,10 +39,16 @@ module Azure::Profiles::Latest
           if(client_0.respond_to?(:subscription_id))
             client_0.subscription_id = configurable.subscription_id
           end
-
+          add_telemetry(client_0)
           @servers = client_0.servers
 
           @model_classes = ModelClasses.new
+        end
+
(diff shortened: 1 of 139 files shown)


azure_sdk/lib/latest/modules/analysisservices_profile_module.rb
```diff
@@ -39,12 +39,18 @@ def initialize(configurable, base_url=nil, options=nil)
...
           end
-
+          add_telemetry(client_0)
           @servers = client_0.servers
 
           @model_classes = ModelClasses.new
         end
 
+        def add_telemetry(client)
+          profile_information = 'Profiles/Rollup/Mgmt/AnalysisServices'
```
veronicagg: is there a reason for us not to follow same order (similar string) as in the namespace, for example record.. "Azure/Profiles/Latest/AnalysisServices/Mgmt"
Does the full telemetry string end up looking something like this: "Profiles/Rollup/Mgmt/AnalysisServices/Latest/azure_mgmt_analysis_services/0,15,2" ? Are we able to infer the api-versions from this user agent? In the case of the person using a profile, I think we could by checking the configuration of the "Latest" profile, in the case of using a specific api-version, would we be able to detect which api-version is being used? I'm trying to look at this overall, so we know telemetry gives us what we need.

sarangan12: Done. I have changed the format. The version is always available from the additional string. So, we do not have to add them here too

generators/profilegen/src/resources/templates/rollup/module_template.template
```diff
@@ -42,6 +43,12 @@ module Azure::Profiles::<%= @profile_name %>
         @model_classes = ModelClasses.new
       end
 
+      def add_telemetry(client)
+        profile_information = 'Profiles/Rollup/<%= @module_name %>'
```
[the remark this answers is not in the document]
sarangan12: Done

azure_sdk/lib/latest/modules/analysisservices_profile_module.rb
veronicagg: I believe you're referring to the version of the gem, I was wondering about the api-version and if what we get overall would be enough for what we want. Thoughts? For profiles itself seems enough, though I believe we'd need to look at the profile configuration for that gem version to know which api-version got picked up, so was wondering if we wanted to consider adding something else.

PR #1282, opened 2018-04-03
Msi Imds task

sarangan12:
@yugangw-msft @salameer Please review and approve

### DIFF
--- a/runtime/ms_rest_azure/lib/ms_rest_azure/credentials/msi_token_provider.rb
+++ b/runtime/ms_rest_azure/lib/ms_rest_azure/credentials/msi_token_provider.rb
@@ -14,6 +14,7 @@ module MsRestAzure
     REQUEST_BODY_PATTERN = 'resource={resource_uri}'
     USER_ASSIGNED_IDENTITY = '{id_type}={user_assigned_identity}'
     DEFAULT_SCHEME = 'Bearer'
+    IMDS_TOKEN_ACQUIRE_URL = 'http://169.254.169.254/metadata/identity/oauth2/token'
 
     # @return [MSIActiveDirectoryServiceSettings] settings.
     attr_accessor :settings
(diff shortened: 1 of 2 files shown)


runtime/ms_rest_azure/lib/ms_rest_azure/credentials/msi_token_provider.rb
```diff
@@ -78,12 +78,68 @@ def initialize(port = 50342, settings = ActiveDirectoryServiceSettings.get_azure
     #
     # @return [String] authentication headers.
     def get_authentication_header
-      acquire_token if token_expired
+      if !ENV['MSI_VM'].nil? && ENV['MSI_VM'].downcase == 'true'
```
yugangw-msft: Wonder what is this environment variable of `MSI_VM` used for? Also is there a in-memory token cache in place to mitigate the throttling issue?

sarangan12: what is this environment variable of MSI_VM used for? 

If the variable is set to true, then the token is obtained using the old method. 

 is there a in-memory token cache in place to mitigate the throttling issue?

Actually Yes. We do not get a new token until the current one is expired.

yugangw-msft: > Actually Yes. We do not get a new token until the current one is expired.

Token is per resource and per identity. If the same provider could be used by multiple clients or multiple identities, then a single token entry might not be enough. I am not familiar with how the `MSITokenProvider` gets used, but hope it helps.

sarangan12: That helps. Actually the same provider will not be used in the Ruby SDK. All of them use their own instance and works fine for all the other token providers.

yugangw-msft: @sarangan12 , sorry for not calling out earlier, but if you have not done it, can you set the `user-agent` header to token requests send to imds. This will help AAD group to use telemetry to diagnose throttling symptoms

PR #1537, opened 2018-07-30
Runtime changes

sarangan12:
@salameer @veronicagg FYI.....


runtime/ms_rest_azure/lib/ms_rest_azure/azure_service_client.rb
```diff
@@ -47,13 +47,22 @@ def get_long_running_operation_result(azure_response, custom_deserialization_blo
             if !polling_state.azure_async_operation_header_link.nil?
               update_state_from_azure_async_operation_header(polling_state.get_request(headers: request.headers, base_uri: request.base_uri, user_agent_extended: user_agent_extended), polling_state)
             elsif !polling_state.location_header_link.nil?
-              update_state_from_location_header(polling_state.get_request(headers: request.headers, base_uri: request.base_uri, user_agent_extended: user_agent_extended), polling_state, custom_deserialization_block)
+              update_state_from_location_header(polling_state.get_request(headers: request.headers, base_uri: request.base_uri, user_agent_extended: user_agent_extended), polling_state, custom_deserialization_block, final_state_via)
```
veronicagg: do we need to be careful with order of parameters due to the deserialization block parameter? should that be the last one? not sure if there are restrictions for blocks - found something here, not sure if it helps https://stackoverflow.com/questions/38614973/ruby-method-with-optional-options-and-block-parameter

runtime/ms_rest_azure/lib/ms_rest_azure/azure_service_client.rb
```diff
@@ -153,7 +168,7 @@ def update_state_from_get_resource_operation(request, polling_state, custom_dese
     # @param polling_state [MsRestAzure::PollingState] polling state to update.
     # @param custom_deserialization_block [Proc] custom deserialization method for parsing response.
     #
```
veronicagg: update comment on parameters

runtime/ms_rest_azure/lib/ms_rest_azure/azure_service_client.rb
```diff
@@ -33,7 +33,7 @@ def initialize(credentials, options = nil)
...
```
veronicagg: update comment for parameters

runtime/ms_rest_azure/lib/ms_rest_azure/azure_service_client.rb
```diff
@@ -153,7 +168,7 @@ def update_state_from_get_resource_operation(request, polling_state, custom_dese
     # @param polling_state [MsRestAzure::PollingState] polling state to update.
     # @param custom_deserialization_block [Proc] custom deserialization method for parsing response.
     #
-    def update_state_from_location_header(request, polling_state, custom_deserialization_block)
+    def update_state_from_location_header(request, polling_state, custom_deserialization_block, final_state_via)
```
veronicagg: should we have a default value for final_state_via here as well?

sarangan12: I think it is better to add a default value here also. Added. Thanks

runtime/ms_rest_azure/lib/ms_rest_azure/azure_service_client.rb
sarangan12: DOne

runtime/ms_rest_azure/lib/ms_rest_azure/azure_service_client.rb
sarangan12: Done

runtime/ms_rest_azure/lib/ms_rest_azure/azure_service_client.rb
sarangan12: Agreed that if it is a block parameter it should be last. But in this case, though it is called custom deserialization block, it is not the traditional block parameter of ruby that is indicated with '&' in method signature. So, adding the optional parameter at the end should be fine (Also it works with the existing code) :-)

veronicagg: sounds good, thx!

PR #1694, opened 2018-09-24
Code change to retry based on retry-after header

sarangan12:
@veronicagg @devigned Please review and approve.

@salameer FYI...

### DIFF
--- a/runtime/ms_rest/lib/ms_rest/http_operation_request.rb
+++ b/runtime/ms_rest/lib/ms_rest/http_operation_request.rb
@@ -83,7 +83,7 @@ module MsRest
             yield(req) if block_given?
           end
 
-          break if (@response.status != 429 && @response.headers['retry-after'].nil?)
+          break if ((@response.status != 429) || (@response.status == 429 && @response.headers['retry-after'].nil?))
 
           if(@response.status == 429 && !@response.headers['retry-after'].nil?)
             sleep(@response.headers['retry-after'].to_i)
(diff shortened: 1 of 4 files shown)


runtime/ms_rest/lib/ms_rest/http_operation_request.rb
```diff
@@ -77,14 +77,22 @@ def run_promise(&block)
...
-          yield(req) if block_given?
+        loop do
+          @response = @connection.run_request(:"#{method}", build_path, body, {'User-Agent' => user_agent}.merge(headers)) do |req|
+            req.params = req.params.merge(query_params.reject{|_, v| v.nil?}) unless query_params.nil?
+            yield(req) if block_given?
+          end
+
+          break if ((@response.status != 429) || (@response.status == 429 && @response.headers['retry-after'].nil?))
+
+          if(@response.status == 429 && !@response.headers['retry-after'].nil?)
```
veronicagg: is this retry behavior general or Azure only? should it belong to ms_rest or ms_rest_azure?

veronicagg:
@sarangan12 Ruby build drop is failing, please check that one before merging.

runtime/ms_rest/lib/ms_rest/http_operation_request.rb
sarangan12: The throttling seems to be general in all other SDKs. So, following the same.

sarangan12:
Ruby Build drop is failing for a known issue. I will fix it soon.

PR #2039, opened 2018-12-04
Support username/passsword OAuth2 authentication

miha-plesko:
With this commit we add a new token provider capable of managing Oauth2 token, but based on username/password combination instead of client_id/client_secret. Example usage:

```
credentials = MsRest::TokenCredentials.new(
  MsRestAzure::PasswordTokenProvider.new(
    'tenant_id',
    'client_id',
    'username',
    'password',
    active_directory_settings
  )
)

options = {
  :subscription_id           => '1111-222-333-444-555',
  :credentials               => credentials,
  :active_directory_settings => active_directory_settings,
}

compute_client = Azure::Compute::Profiles::Latest::Mgmt::Client.new(options)
...
```

This kind of token management is required by Azure Stack where client_id is hardcoded without accompanying secret. See

https://docs.microsoft.com/en-us/azure/azure-stack/user/azure-stack-rest-api-use#example-1


miha-plesko:
@bsiegel or @sarangan12 looking forward to your review.

bsiegel:
@sarangan12 Any update here?

sarangan12:
This has been opened for quite sometime. A few comments:

1. I do not think this username & password is a requirement in using our SDK. No customer has specifically asked for such a request. 

2. Even if it is required, I do not think this is sufficient. Because, the passwords cannot and should not be just serialized and sent it wire without encryption. 

I am closing this PR now. If you think there is a legitimate customer requirement, feel free to reopen

miha-plesko:
@sarangan12 as far as I'm reading the documentation, username/password authentication seems the only way to authenticate against AzureStack API, see [here](https://docs.microsoft.com/en-us/azure-stack/user/azure-stack-rest-api-use#example-1). Can you please provide an example of how to authenticate in any other way?

Fryguy:
@sarangan12 @bsiegel  Any update for @miha-plesko 's question here?

bsiegel:
@yungezz @MyronFanQiu are the maintainers these days, they should be able to assist with this request.

@Fryguy I see that you have an implementation of the requested logic, would you consider opening a PR against this repo? That would likely speed this up, as otherwise I suspect this issue will hang out in the backlog until the team prioritizes it up.

miha-plesko:
@bsiegel but this *is* a PR :)

miha-plesko:
@sijuman this is the PR to support username/password authentication against AzureStack (as discussed on the call earlier today). We're currently using it in CloudForms provider by applying it as a monkey-patch see here: https://github.com/manageiq/manageiq-providers-azure_stack/blob/master/app/models/manageiq/providers/azure_stack/manager_mixin.rb#L88. Would be nicer to have it reviewed and merged upstream :)

/cc @mancabizjak @gberginc

PR #2040, opened 2018-12-04
Let Travis also test against Ruby v2.4.4

miha-plesko:
With this commit we make Travis also test against the same Ruby version as [ManageIQ](https://github.com/ManageIQ/manageiq) project is targeting to ensure they are compatible.


bsiegel:
@sarangan12 Any update here?

PR #2398, opened 2019-04-13
Azure CLI Token provider

### LINKED ISSUE #2381: CLI Token Provider support
The .NET package for MSI authentication within App Services includes a default fallback to retrieve an Azure CLI access token if no Managed Service Identity is available.  This is useful as now no keys need to be used at all both in dev and production.  CLI context is used locally and MSI context when deployed.

It would be beneficial to add a CLITokenProvider class that sets all necessary options so the same approach can be taken from a Ruby application.

The C# implementation is available here:

https://github.com/Azure/azure-sdk-for-net/blob/40248f466aa063bd90f16dae10d06b9b7317e2d0/src/SdkCommon/AppAuthentication/Azure.Services.AppAuthentication/TokenProviders/AzureCliAccessTokenProvider.cs

jlorich:
Resolves #2381

This adds an `AzureCliTokenProvider` class which leverages the locally installed azure cli to generate an access token on behalf of the currently logged in user.  I find this to be an extremely helpful feature specifically because it allows keys to be fully extracted away both in local testing and when deployed to an environment featuring a Managed Service Identity

 This has already been implemented for the C# SDK, so this should bring things up to parity.


jlorich:
@sarangan12 Any feedback? I've got a customer who'd like to use this. Tests are in place and passing and I've tried it out myself in a few environments.

jlorich:
bump

jlorich:
This is reviewed tested and approved.  Can we get a merge?

PR #2423, opened 2019-04-25
2019 stack profile

sarangan12:
@veronicagg @salameer FYI....


config/profiles.json
```diff
@@ -1,6 +1,82 @@
...
+                },
+                "Microsoft.Subscriptions" : {
+                    "management": {
+                        "2016-06-01": ["*"]
+                    }
+                },
+                "Microsoft.Storage" : {
+                    "management": {
+                        "2017-10-01": ["*"]
+                    }
```
bganapa: I believe storage data plane still resides in a separate repo and its hand crafted..
We use the following packages for AZURESTACK at the moment, I will see whether we need to take any later versions
https://rubygems.org/gems/azure-storage-common/versions/1.0.1
Blob: https://rubygems.org/gems/azure-storage-blob/versions/1.0.1
Queue: https://rubygems.org/gems/azure-storage-queue/versions/1.0.1
Table: https://rubygems.org/gems/azure-storage-table/versions/1.0.1

azure_sdk/lib/azure_sdk.rb
```diff
@@ -6,3 +6,4 @@
 require 'v2017_03_09/v2017_03_09_profile_client'
 require 'latest/latest_profile_client'
 require 'v2018_03_01/v2018_03_01_profile_client'
+require 'v2019_03_01/v2019_03_01_profile_client'
```
bganapa: Soon we will have profiles for government clouds which will be called with -Profile rather than -Hybrid. The Hybrid profiles are for AZURESTACK and for all the public/gov clouds. Meaning the app developed with -Hybrid will work across all the clouds AZURESTACK, UsGov, AzureChina and AzureGermany..

I am thinking whether we should call the client as v2019_03_01_hybrid instead of v2019_03_01_profile.. I am checking with Vishnu on the other SDKs usage pattern. I understand we will deviate from the 2018 profile..

azure_sdk/lib/v2019_03_01/modules/authorization_profile_module.rb
```diff
@@ -0,0 +1,133 @@
+# encoding: utf-8
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+
+require 'azure_mgmt_authorization'
+
+module Azure::Profiles::V2019_03_01
```
bganapa: Here Azure::Profiles::V2019_03_01 should be Azure::Profiles::V2019_03_01_hybrid ? Checkign with Vishnu on the usage of other SDKs

azure_sdk/lib/azure_sdk.rb
bganapa: We have the _hybrid for Java/C#/Python SDKs, but not for Go and Ruby. @sarangan12 going forward can we call this with _hybrid

azure_sdk/lib/v2019_03_01/modules/authorization_profile_module.rb
bganapa: We have the _hybrid for Java/C#/Python SDKs, but not for Go and Ruby. @sarangan12 going forward can we append with _hybrid

azure_sdk/lib/azure_sdk.rb
sarangan12: Done

azure_sdk/lib/v2019_03_01/modules/authorization_profile_module.rb
sarangan12: Done

Issue #2499, opened 2019-05-28
UsageDetail attribute pretax_cost has wrong type.


aheumaier:
In Module[ Azure::Consumption::Mgmt the class UsageDetail](https://github.com/Azure/azure-sdk-for-ruby/blob/master/management/azure_mgmt_consumption/lib/2018-06-30/generated/azure_mgmt_consumption/models/usage_detail.rb)  uses an _Integer class_ for attribute :pretax_cost. This results in values of 0  - while  most of the values are in the form _Float_ as 0.0005.
This field has to be a float for accurate cost calculations (correctly implemented in Python SDK)

```
# Poor mans monthly cost calculator
#
response = Azure::Consumption::Profiles::Latest::Mgmt::Client.new.usage_details.list(filter: "properties/usageEnd ge '2019-04-01' AND properties/usageEnd le '2019-04-30'")

sum = 0.0

response.each do |item|
  sum += item.pretax_cost
end

puts sum

```

Please let me know if you have any questions - happy to help!

sgiacomel:
This applies also to these properties:
usageQuantity
billableQuantity

sgiacomel:
Reported here: 
https://social.msdn.microsoft.com/Forums/azure/en-US/266d4123-3407-4e65-8f79-09a64ac857bd/bug-in-azuresdkforruby-consumption?forum=windowsazuredevelopment

kurtzeborn:
Thank you for your interest in Azure SDKs.  As detailed in this [retirement announcement](https://github.com/Azure/azure-sdk-for-ruby#important-announcement), this repo is no longer supported as of December 31st 2021.  Please find the up-to-date list of languages and services supported with Azure SDKs here: https://aka.ms/azsdk

PR #2694, opened 2019-06-28
Fix: Changed azure_mgmt_consumption UsageDetail::pretax_cost type to …

aheumaier:
Fix for #2499.  Changed azure_mgmt_consumption UsageDetail::pretax_cost type to Double


sgiacomel:
This should be applied also to these properties:
usageQuantity
billableQuantity

PR #2733, opened 2019-11-20
Fix #2732 - [De]Serialize TimeSpan type properly

### LINKED ISSUE #2732: Serialization of TimeSpan type produces empty hash
I'm trying to create an autoscale rule for a VM Scale Set, using the Azure::Monitor::Mgmt::V2015_04_01 module, but I'm getting error responses with the body: `{"code":"UnsupportedRequestContent","message":"Request content is not well formed or supported."}`. Upon further inspection, my properties with the TimeSpan type are getting converted to an empty hash instead of a string value upon serialization.

Here is some code to duplicate the bug:

```ruby
action = Azure::Monitor::Mgmt::V2015_04_01::Models::ScaleAction.new
action.cooldown = 'PT10M'
action.direction = 'Increase'
action.type = 'PercentChangeCount'

request_mapper = Azure::Monitor::Mgmt::V2015_04_01::Models::ScaleAction.mapper()
monitor = Azure::Monitor::Profiles::Latest::Mgmt::Client.new
monitor.autoscale_settings.client.serialize(request_mapper, action)
```

This code produces the following result:

```ruby
{"direction"=>"Increase", "type"=>"PercentChangeCount", "value"=>"1", "cooldown"=>{}}
```

### Why

I dug into the [serialize method](https://github.com/Azure/azure-sdk-for-ruby/blob/master/runtime/ms_rest/lib/ms_rest/serialization.rb#L240) and noticed that there is no matching regex for a `TimeSpan` type, as referred to [here](https://github.com/Azure/azure-sdk-for-ruby/blob/master/management/azure_mgmt_monitor/lib/2015-04-01/generated/azure_mgmt_monitor/models/scale_action.rb#L81) and lots of other places. I believe that either this method needs updated to support the `TimeSpan` type, or every reference to a `TimeSpan` type needs changed to a `String` type.

jsmartt:
Fixes #2732


bim-msft:
@jsmartt Thanks for your contribution, we'll look into this.

PR #2741, opened 2019-12-11
Release 20191212

bim-msft:
- [ ] azure_mgmt_container_service 0.19.0 -> 0.20.0
- [ ] azure_mgmt_deployment_manager 0.17.0
- [ ] azure_mgmt_iot_central 0.17.3 -> 0.18.0
- [ ] azure_mgmt_policy_insights 0.17.4 -> 0.17.5

- [ ] azure_sdk 0.47.1 -> 0.48.0


config/profiles.json
```diff
@@ -1293,10 +1298,9 @@
                 "Microsoft.ContainerService" : {
                     "management": {
                         "2017-07-01": ["*"],
-                        "2019-04-01": ["*"],
```
MyronFanQiu: could you provide the context for removing this version?

bim-msft: @MyronFanQiu The latest version here doesn't contain version `2019-04-01`, so I removed the version in our profiles.json, is this the correct modification? Or I should keep all historical versions? 
https://github.com/Azure/azure-rest-api-specs/tree/c1aecc73fc025d3db2714717f69c3560ad23e64d/specification/containerservice/resource-manager#tag-package-2019-11

MyronFanQiu: It's not the same with the latest tag in swagger spec. In my understanding, we should not remove the old api. But we do need a way to decide which version should be added or not.

bim-msft: @MyronFanQiu Thanks, this piece of code has been reverted and regenerated, please review it again.

PR #2754, opened 2020-02-04
Update faraday version to allow 1.0.0

d-m-u:
[Faraday got updated](https://github.com/lostisland/faraday/pull/1101) and [we over at manageiq](https://github.com/ManageIQ/manageiq/issues/19760) were hoping maybe that you all might be amenable to letting people update so we can support ruby 2.7.

Please let me know if anything else is needed.

Thanks!


bim-msft:
@d-m-u Thanks for the information.
@MyronFanQiu We should release new version of `ms_rest` and `ms_rest_azure` for this.

MyronFanQiu:
@bim-msft Sure. @d-m-u Sorry, due to the coronavirus, we have to work from home until next week. I think we can merge this PR and release a new version next week.

d-m-u:
@bim-msft and @MyronFanQiu thank you!

PR #2771, opened 2020-04-23
Release 20200422

MyronFanQiu:


config/profiles.json
```diff
@@ -2107,7 +2126,7 @@
                     "management": {
                         "2017-03-30": ["*"],
                         "2019-06-01": ["*"],
-                        "2019-12-01": ["*"]
+                        "2020-03-01": ["*"]
```
bim-msft, line 2129: Hi, I have a question: why replace the old version here? Is it ok to support `2019-12-01` and `2020-03-01` at the same time?

MyronFanQiu: Based on my understanding, the profile should use the latest version. For example, if service team changed the name of the properties, including both api-version would cause problem. But your question and concern is correct. It depends on how we define the word `latest`. For some sub resources within network RP, 2019-06-01 is the latest version for them. However, for the whole network PR, the latest version is 2020-03-01. Actually, we cannot know whether the sub resources are deprecated intentionally or not.

PR #2782, opened 2020-05-21
update aad endpoint in US cloud

### LINKED ISSUE #2781: Azure Gov environment uses an out-of-date default for the active directory endpoint url
The AzureUsGovernment environment embedded within the most recent version of the `ms_rest_azure` gem is currently set to `https://login-us.microsoftonline.com/`

https://github.com/Azure/azure-sdk-for-ruby/blob/master/runtime/ms_rest_azure/lib/ms_rest_azure/azure_environment.rb#L133

However, per this article, the Azure Gov environment should now be using `https://login.microsoftonline.us` instead:
https://devblogs.microsoft.com/azuregov/azure-government-aad-authority-endpoint-update/

This agrees with the Azure Government developer documentation on endpoint mapping (see the Active Directory Endpoint and Authority entry):
https://docs.microsoft.com/en-us/azure/azure-government/documentation-government-developer-guide#endpoint-mapping

It appears that the old `https://login-us.microsoftonline.com/` endpoint has finally been deprecated within the Azure Gov environment. Within the last week multiple customers have been reporting that code using the `azure_mgmt_compute` gem (which we believe uses the `ms_rest_azure` gem internally) has starting failing when connecting.

Would it be possible to get the default active directory endpoint updated to the new `https://login.microsoftonline.us` URL per Azure Gov documentation?

MyronFanQiu:
Fix #2781

### DIFF
--- a/runtime/ms_rest_azure/lib/ms_rest_azure/azure_environment.rb
+++ b/runtime/ms_rest_azure/lib/ms_rest_azure/azure_environment.rb
@@ -130,7 +130,7 @@ module MsRestAzure
                                                         :sql_management_endpoint_url => 'https://management.core.usgovcloudapi.net:8443/',
                                                         :sql_server_hostname_suffix => '.database.usgovcloudapi.net',
                                                         :gallery_endpoint_url => 'https://gallery.usgovcloudapi.net/',
-                                                        :active_directory_endpoint_url => 'https://login-us.microsoftonline.com/',
+                                                        :active_directory_endpoint_url => 'https://login.microsoftonline.us/',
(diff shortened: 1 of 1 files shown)


MyronFanQiu:
> Hi,
> When would this be merged as we are facing issue in test-kitchen testing.
> We couldn’t login to azure us government cloud.

This issue should has been fixed.

Vijay8888:
Not sure we are still getting error with azureusgovernment environment in test Kitchen kitchen-azurerm . Error as below.
Could not connect to azure. Please check tenant id, client id, client secrete.

PR #2784, opened 2020-05-21
Code upgrade to bump ms rest azure dependency

MyronFanQiu:


danielcbright:
hello fine folks! I'm curious when the azure_keyvault gem that has been bumped will be available on rubygems.org? I'm working on a Chef Cookbook that I'd like to test with it

MyronFanQiu:
> hello fine folks! I'm curious when the azure_keyvault gem that has been bumped will be available on rubygems.org? I'm working on a Chef Cookbook that I'd like to test with it

Hi. They would be released in next week if anything fine.

PR #2810, opened 2020-11-11
key_vault: Fix naming errors in create_certificate_async

JelteF:
`create_certificate_async` was failing because of naming of some types and
attributes. This caused certificate creation not to work using the API,
because it failed before even sending any request.

I'm not sure how the auto generation of this repo works, so I only fixed this 
for v7.1. It seems this bug also exists in other versions of the `azure_key_vault` 
API though.


JelteF:
After looking a bit more into this I think basically none of the `azure_key_vault` methods work. For two reasons:
1. the `XXXParameters` classes should all be prefixed with `Models::`, otherwise their path is incorrect and the class does not exist.
2. if parameters are given these are then assigned to `CamelCase` attributes, but they should use `snake_case`

kairu-ms (approved):

kairu-ms:
Fixed in 0.19.0

PR #2829, opened 2021-03-12
Fix key vault issues

kairu-ms:


serprex:
There's still issues in 0.19.0,
```
NameError: uninitialized constant Azure::KeyVault::V7_2_preview::KeyVaultClient::CertificateIssuerSetParameters
from /home/pdube/.rbenv/versions/2.6.7/lib/ruby/gems/2.6.0/gems/azure_key_vault-0.19.0/lib/7.2-preview/generated/azure_key_vault/key_vault_client.rb:858:in `set_certificate_issuer_async'
```
(Also happens with `V7_1`)

PR #2842, opened 2021-08-24
use require_relative for relative paths

kbrock:
`require` loads files based upon a load path.
`require_relative` loads files based upon the location of the current file.

While `..` is discouraged from use in `require_relative`, it just doesn't make sense in terms of `require` and a load path.

I found this issue when pointing my `Gemfile` at GitHub and my ruby couldn’t make sense of the gemspecs.

If there is a template to update instead, just point me in that direction.

thanks


ArthurMa1978:
Thank you @kbrock , this repo [has entered a retirement phase and is no longer officially supported by Microsoft](https://github.com/Azure/azure-sdk-for-ruby#important-announcement).

kurtzeborn:
Thank you for your interest in Azure SDKs.  As detailed in this [retirement announcement](https://github.com/Azure/azure-sdk-for-ruby#important-announcement), this repo is no longer supported as of December 31st 2021.  Please find the up-to-date list of languages and services supported with Azure SDKs here: https://aka.ms/azsdk

PR #2846, opened 2022-04-25
Bump thor dependency to 1.2

mark-davenport-fountain:
This will allow us to update some other projects in our rails application such as bundler-audit which requires a newer version of Thor. 

The only impacted file in the repo is `bin/pfxer` which is not effected by updating to Thor 1.2. 

I do not have a publish settings file to test with, but I made a mock XML file and confirmed that the output was the same before and after the upgrade to Thor.


mark-davenport-fountain:
I somehow doubt the `soon` part @msftdata

